PR #1, opened 2023-10-10
Small Lumber Rework

CodingBlueFrogger:
First time messing with KubeJs and saw simple work that could be done on an anticipated pack

- Ported cutter +assembler recipes from 1.12.2 (not including gate recipe)

- Added gt saw compat for tfc lumber recipes (Probably to be removed at later date when saws are merged)

- Added create saw recipes based on horse drawn cutting block from 1.12.2 + create's log stripping

### DIFF
--- a/kubejs/server_scripts/common/wood_names.js
+++ b/kubejs/server_scripts/common/wood_names.js
@@ -1,0 +1,24 @@
+// priority: 0
+
+const WOOD_NAMES = [
+  "acacia",
+  "ash",
+  "aspen",
+  "birch",
+  "blackwood",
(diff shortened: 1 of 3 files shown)


SpeeeDCraft:
Looks good, but i think firstly we need to fix GT oregen in TFC world

SpeeeDCraft (approved):

CodingBlueFrogger:
Closing PR as I've rewritten a decent amount since this PR initially opened. Better realized how item tags work and plan to use recent commits which have better locations for constant values 👍

Issue #364, opened 2024-05-20
sophisticated backpacks not dye-able


caelansimms:
### Version modpack Used

0.7.8

### Environment

Singleplayer

### New Worlds

Yes

### Expected Behavior

Backpacks supplied from SophisticatedBackpacks Mod are dye-able, involving the body of the backpack and the straps of the backpack, allowing for 2 colors to be added on

### Actual Behavior

Backpacks are not able to be dyed with any dye inside a crafting menu, 2x2 or 3x3.

### Steps to Reproduce

1. Craft/grab a basic backpack (tier 1 or otherwise)
2. Select any dye and attempt to dye the backpack by placing a dye/dyes inside the grid alongside the backpack 

### Additional Information

This _is_ a feature inside other versions and other modpacks, with no addons needed to make this possible. I highly doubt that it was removed. I can only suspect in the process of reworking the backpacks outside of "Backpack Tier X" catagories, substituted with the present Backpack, Red Steel Backpack, Titanium Backpack, etc. that this is where the hiccup originates.
Cheers.

Exzept1on:
Works right.

Exzept1on:
Thanks for shit dev, who can't fix bugs or problems with his mod.

Exzept1on:
Maybe we will fix this by install other backpack mod

caelansimms:
Okay, didn’t know it was a mod-specific bug. 

Personally, this is IMO the best backpack mod going around & the bug is purely cosmetic so I can live with it, I request we don’t change it.

thanks for replying!

BabBimmaBome:
welcome to trying to use tfc and greg with other mods stuff isnt going to work right and the mod devs dont help lol

> Okay, didn’t know it was a mod-specific bug.
> 
> Personally, this is IMO the best backpack mod going around & the bug is purely cosmetic so I can live with it, I request we don’t change it.
> 
> thanks for replying!
welcome to trying to use tfc and greg with other mods stuff isnt going to work right and the devs for these mods dont help lol

BabBimmaBome:
https://www.curseforge.com/minecraft/mc-mods/useful-backpacks maybe this oldy but a goodie would work?

Exzept1on:
Yes, it was in the direction of this mod that we looked, we also considered backpacks from crayfish, but have not yet decided.

Exzept1on:
> Okay, didn’t know it was a mod-specific bug.
> 
[...]

PR #378, opened 2024-06-01
Temporal fix for #364

chemlzh:
## What is it?
Temporal fix for #364, now backpacks can be dyed, both belt and pack. However, GTM chemical dyes can also be used for recipes, which is not expected. Besides, backpacks cannot be dyed in chemical bath machines. The difficulty to fix these bugs is that I have not figured out how to set dyed color, since it's not replacing but mixing with original color.


chemlzh:
@Xikaro @FriendlyCube Do you have any idea for this (temporal) backpack dyeing fix?

Exzept1on:
I'll think about this fix myself later, I'm a little busy right now

Issue #586, opened 2024-11-04
Adding Issue Labels for Language Filtering


Supermarcel10:
### Version modpack Used

N/A

### Related Problem

Filtering issues is difficult in terms of languages.

### Your Solution

Adding language labels like:
`en` and `ru`

cdecompilador:
Maybe an autotranslate bot to english would be convenient

PR #587, opened 2024-11-04
- Added labels to issue templates.

Supermarcel10:
## What is Implemented
Solution to make filtering issues and pull requests easier by language.
Solution for #586.

## Implementation Details
Updated issue templates to contain language issue labels.

> [!NOTE]
> Labels still have to be created manually via GH!


Xikaro:
I think it's worth reconsidering this and sticking with EN version only

Xikaro:
Thanks for the PR, I changed the issue template, only English versions remain

PR #634, opened 2024-12-18
TFC Stone Hammer Restoration

ThePatmann:
## What
Restores the recipes for the Terrafirmacraft Stone Hammer, undoing the replacement with gregtech's stone hammer. This is necessary as the gregtech stone hammer allows players to skip terrafirmacraft progression by mining with it as soon as they can knap stones together.

## Implementation Details
Before Pulling, verify that changes retain full gameplay functionality of the hammer while preventing players from mining ore before casting their first pickaxe.

## Outcome
Disables GTCEU stone hammer via disabled-items constant list. 
Removes TFC stone hammer from disabled-items constant list, restoring its tags and functionality
Removes TFC stone hammer recipe interference/replacement from TFC recipe.js, returning the original recipe to the game and removing the only way to craft the GTCEU stone hammer (as it is no longer necessary)

Save files which already have Gregtech Stone Hammers will find them unusable and unobtainable, but not truly removed from the game.

### DIFF
--- a/kubejs/server_scripts/tfc/recipes.js
+++ b/kubejs/server_scripts/tfc/recipes.js
@@ -2759,31 +2759,6 @@ const registerTFCRecipes = (event) => {
 
     //#endregion
 
-    //#region Молот
-
-    // Инструмент
-    event.remove({ id: `tfc:crafting/stone/hammer_igneous_extrusive` })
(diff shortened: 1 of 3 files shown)


ThePatmann:
Wrong Repo!

PR #637, opened 2024-12-19
Fixed oilsands, added missing english strings, added indicators for all ores

Pyritie:
image
image
image
image


Pyritie:
I wanted to separate the strings from the ore indication change but idk how to github properly. Feel free to reject that while accepting the strings, or whichever you like.

Anyway, I added ore indicators to all veins:
* Redstone and graphite have pretty deep indicators, similar to the gold/copper/iron/etc ones that already exist, since they are required for progression
* Everything else has shallow indicators, not much further than a prop pick, since they are not very relevant
* Redstone wire, salt licks, amethyst clusters, and certus clusters are now also indicators
* I only use the gregtech indicators if there's nothing else better, because they're kinda ugly
* I added a lot more indicators underground to make the caves more interesting, because otherwise they are very empty

I also adjusted the size of naquadah and tetrahedrite after some very late game players complained about them.

Please let me know what you think!

image
image
image
image

Pyritie:
idk how to mark this as ready for review

Exzept1on:
What version of greg you are use? (in this pr)

Pyritie:
The one in 0.7.14, I have not changed greg version

PR #639, opened 2024-12-22
Fixed not being able to make cobble in assembler

Pyritie:
Added circuit number to the rock -> cobble recipe in assembler.

Before it was impossible because it would always create bricks.


SverhRazum-Nah:
image
image
In the new version, concrete is needed there (Just like in the original)
image

PR #643, opened 2024-12-22
Fix casing exploit, change recipes for tfc alloys

Pyritie:
Changed the LV casing assembler recipe so it always requires colored steel. This matches the behaviour in 0.8.1.
This also stops exploits like getting infinite iron from dusts and then turning it into other metals through the arc furnace.
image

I have also changed the other gregtech equivalents of tfc alloys, so they are easier to craft by machine.
Black bronze, bismuth bronze, sterling silver, and rose gold dust now only need a steam/electric furnace.
Black/red/blue steel need an arc furnace, so they are a bit later in LV.
Normal steel is unchanged.
This way, players can stop using their TFC blast furnace and crucible once they're in LV.

Please let me know what you think!


fluffle:
I think I'd prefer it if you just made the arc furnace / macerator return 8 steel dust -- being able to make LV machine casings from steel once you get an LV assembler is quite a nice leg up in mid LV. Just my $0.02 :-)

Pyritie:
True, but adding arc furnace + macerator recipes for every single machine sounds like a hassle lol - from what I can tell, they're auto generated from whatever the normal crafting recipe is

Pyritie:
Also, I'm just trying not to step on except1on's toes too much -- if I was designing this, I'd do something like only make lv ones need black steel, and have mv need the coloreds instead, because right now mv hulls feel way too easy compared to lv ones

Pyritie:
Closing because #668 was accepted instead.

Exzept1on:
I'll add a little bit to this in the future and push it.

Pyritie:
alright!

Pyritie:
> I'll add a little bit to this in the future and push it.

If it won't be soon, is it something I could do for you?

Exzept1on:
I wanted to add something here, but I have a question, is it ready? That is, I would like the LV casings to be made of red and blue steel and the machines that are disassembled after that, you can get red and blue steel from them, is it possible that you can do this, or has it already been done?

Pyritie:
No, we will have to undo #678 and #668. I did not do that yet in case there was something else you wanted to do too.

Pyritie:
I think I might've already stopped those when I did an earlier merge 😅
I tested this branch in game. It all looks correct!
Ok to merge 👍

PR #644, opened 2024-12-22
Fix stone dust tag

Pyritie:
Fixes the #stone_dusts tag because it was looking for dusts in the gtceu namespace when most of them are in tfg.

image


Pyritie:
I closed this by accident sorry

Issue #636, opened 2024-12-19
[🐛]: Create Crafts & Additions rolling mill makes rods out of rods and doesn't accept ingots.


Jipaso:
### Version modpack Used

0.8.1

### Environment

Singleplayer

### New Worlds

Yes

### Expected Behavior

Expected to produce 2 metal rods by consuming 1 metal ingot.

### Actual Behavior

The mill turns 1 rod into 1 rod of the same type.
![Screenshot 2024-12-19 163225](https://github.com/user-attachments/assets/3d19f933-6011-4f97-a7b0-2c16433d1061)

### Steps to Reproduce

I tried using the rolling mill and the recipes work as shown in EMI.

AlertingAvian:
Rod input can be fixed by modifying recipes.js `ingotItem` should have `tagPrefix.ingot` instead of just `tagPrefix` (line 151 kubejs/server_scripts/gregtech/recipies.js)

Output quantity can be fixed by added a line at around 156 that is `rodItem.count = 2`

```javascript
const processRod = (tagPrefix, material) => {
	const rodItem = ChemicalHelper.get(tagPrefix, material, 1)
	if (rodItem.isEmpty()) return
	
	const ingotItem = ChemicalHelper.get(tagPrefix.ingot, material, 1) // this line needs to have tagPrefix.ingot not just tagPrefix
	if (ingotItem.isEmpty()) return

	if (!material.hasFlag(MaterialFlags.GENERATE_ROD) || material == GTMaterials.Wood) return
	if (ingotItem.isEmpty() || rodItem.isEmpty()) return
	
	rodItem.count = 2
	
	// Прокатка стержней
	e.custom({
		type: "createaddition:rolling",
		input: ingotItem.toJson(),
		result: rodItem.toJson()
	}).id(`tfg:rolling/${material.getName()}_rod`)
}
```

Pyritie:
you can submit a pull request, you can edit files through the github website, that's how I've been doing it

PR #650, opened 2024-12-27
fix rod rolling output amount

AlertingAvian:
## What is the new behavior?

Fixes issue #636, sets output amount of rod rolling recipes to 2

## Implementation Details

Changed output of rolling mill rod recipes to be consistent with 0.7.14

## Outcome

Fixes: #636

### DIFF
--- a/kubejs/server_scripts/gregtech/recipes.js
+++ b/kubejs/server_scripts/gregtech/recipes.js
@@ -145,7 +145,7 @@ const registerGregTechRecipes = (e) => {
     }
 
     const processRod = (tagPrefix, material) => {
-        const rodItem = ChemicalHelper.get(tagPrefix, material, 1)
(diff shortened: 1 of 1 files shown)


SverhRazum-Nah:
You fixed the number. But did you fix the recipe yourself (so that the rods are created from ingots, and not from rods)?

AlertingAvian:
The recipe fix had already been completed by Exzept1on with https://github.com/TerraFirmaGreg-Team/Modpack-Modern/commit/3a62265b5346f688d62b746435593ff6563741e1, I only changed the output amount. The fix for rod input is not in the pre-release though so I didn't see it originally when writing my comment on 636

Pyritie:
Closing because the rolling mill will produce plates and foils in 0.9.

PR #657, opened 2024-12-29
Reverted old ore generation algorithm and applied the latest changes from the main branch

venglov:
## What is the new behavior?
This PR reverts the new ore generation algorithm due to its current issues, which cannot be resolved on our side. Additionally, the latest changes and fixes from the main branch have been merged into this branch, with all conflicts resolved.

## Implementation Details

-  Reverted the new ore generation algorithm because it lacks the necessary configurability to fix its issues, such as setting an offset for layers in the GTCE algorithm. Implementing a fix would require forking and maintaining a custom version of GTCE, which is not ideal.

- Merged the latest changes from the main branch including enhancements for the old generation algorithm introduced by @Pyritie

No significant compatibility issues are expected

I am shitcommitter, please squash & merge :)


Pyritie:
0.9 keeps the same TFC-based ore generation, so this is no longer needed. Thank you!

PR #704, opened 2025-01-13
Added recipes for create glass, more windows, fixed tinted glass recipe conflict

Pyritie:
## What is the new behavior?
There's now more glass

image

image
image
image
image
image
image

All of these have pane recipes that matches the existing ones
image
image

Also, now there's windows for every wood type
image

Also also, I fixed tinted glass because it used black dye before, and that makes black stained glass. Now it's amethyst like the glassblowing recipe
image

## Implementation Details
Requires Every Compat https://www.curseforge.com/minecraft/mc-mods/every-compat
Which also wanted moonlight lib to be updated. (I did and haven't had any problems? I haven't tested much besides these glass recipes though)

## Outcome
More building blocks is always good, and the vanilla glass is ugly by itself


Pyritie:
also I can't create a pull request for "nothing", but please remove Inventory Sorter and add https://www.curseforge.com/minecraft/mc-mods/inventory-tweaks-refoxed (specifically v1.1) instead please, it lets you sort inside greg crates!

PR #708, opened 2025-01-13
Functional storage for 0.7

SverhRazum-Nah:
Функциональное хранилище для версии модпака 0.7. Изменены крафты и всё такое. Вы и так знаете
Functional storage for modpack version 0.7. Changed crafting and all that. You already know.

Моды необходимые для работы:
Mods required for your work:
[mods.zip](https://github.com/user-attachments/files/18397830/mods.zip)

Так же можно удалить Storage Drawers. Всё равно у них нет рецептов
You can also delete Storage Drawers. They don't have any recipes anyway


Pyritie:
does this have similar balance to the backpacks? like only the first stack upgrade being available etc... some of the higher tier ones get really crazy and basically make any other kind of storage useless

SverhRazum-Nah:
> does this have similar balance to the backpacks? like only the first stack upgrade being available etc... some of the higher tier ones get really crazy and basically make any other kind of storage useless

I know that the upgrades are quite strong, so I made their recipes the same as those of the super chests:
image
image


image
image


image
image

So on lore site it's like you're adding super chest technology to the drawers.

And as for me, they do not change the logic of progression, but only complement it:

Lots of items with an average number of each -> Drawers
There are few items but each of them has a large number -> it is much cheaper to make just a super chest
Once you get ME system and storage buses, you don't need drovers as much, but you can still connect them to the system to store an average number of identical items.

Pyritie:
how many items can each stack hold in the fully upgraded one?

Pyritie:
@Xikaro This one will wait until after 0.9 is stable, we can add it whenever we do the Create 6 update

Pyritie:
@SverhRazum-Nah Sorry, but I think we'll end up not using this, just because Create 6 kinda solved the problem in a different way, and a mod like [this](https://modrinth.com/mod/create_factory_logistics) would fit into its system better.

SverhRazum-Nah:
> @SverhRazum-Nah Sorry, but I think we'll end up not using this, just because Create 6 kinda solved the problem in a different way, and a mod like [this](https://modrinth.com/mod/create_factory_logistics) would fit into its system better.

NOOOOOOOOOOOOOOOOOOOOOOOOOOOOO  (╥_╥)(╥_╥)(╥_╥)

Pyritie:
sorryyyyyy

PR #713, opened 2025-01-13
Allows all colors of glass to be macerated/hammered to dust

CaptainGold1:
## What is the new behavior?
All colors of glass (and tinted glass as well) can now be macerated in the macerator or hammered in the crafting window to dust. Colored glass panes work as well.

## Implementation Details
The default maceration recipes for glass and glass panes were removed and replaced with #forge:glass and #forge:glass_panes.

## Outcome
Most TFC sands by default output colored glass, while previously only clear glass could provide glass dust (needed for vacuum tubes). This means you had to go out searching for white sand for silica batches to progress into LV. Allowing colored glasses allows all sand types to be used for glass dust.

## Additional Information
The Create bamboo and cherry glass panes are tagged with #forge:glass_panes, so show up in the list of possible inputs for tiny glass dust. However, these items are hidden from EMI and have no recipes, so they're not intended inputs. A possible fix is to remove #forge:glass_panes from these blocks?

### DIFF
--- a/kubejs/server_scripts/create_connected/tags.js
+++ b/kubejs/server_scripts/create_connected/tags.js
@@ -1,0 +1,7 @@
+const registerCreateConnectedItemTags = (event) => {
+    // Removes tags from the Cherry and Bamboo wooden panes added by Create Connected
+    event.removeAllTagsFrom("create_connected:cherry_window_pane")
+    event.add("c:hidden_from_recipe_viewers", "create_connected:cherry_window_pane")
+    event.removeAllTagsFrom("create_connected:bamboo_window_pane")
+    event.add("c:hidden_from_recipe_viewers", "create_connected:bamboo_window_pane")
+}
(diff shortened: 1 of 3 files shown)


Pyritie:
> The Create bamboo and cherry glass panes are tagged with #forge:glass_panes, so show up in the list of possible inputs for tiny glass dust. However, these items are hidden from EMI and have no recipes, so they're not intended inputs. A possible fix is to remove #forge:glass_panes from these blocks?
If they aren't craftable, does adding `c:hidden_from_recipe_viewers` to those items work?

CaptainGold1:
Probably, I can go ahead and test that in a couple hours.

CaptainGold1:
Okay, I committed a change that seems to work. It seems that those specific glass panes were added by create_connected and thus aren't in the create blacklist. I mirrored the implementation that blacklists create items into a new js file, hopefully it's the proper way to do it but point out if it isn't. The glass panes no longer show up as valid inputs.

PR #717, opened 2025-01-15
Create Sails: Recipe Improvements

Nebby1999:
## What is the new behavior?

This PR looks to make the utilization of Sail blocks more accesible by doing the following changes:

1. Utilizing Burlap or Wool cloth instead of blocks of Wool (cuts down steps alongside giving access to secondary means of sail creation.
2. Giving a proper crafting recipe for Create's Sail Frames, by utilizing lumber and wrought iron screws
3. Nearly doubling the efficiency of the sail creation by making 1 cloth give out 8 sails, instead of 1 cloth giving a total of 4 sails at the cost of using lumber instead of sticks
4. This makes windmills slightly more viable source of SU before you can create other SU generating methods
5. Also allows access to windmills without relying on farm animals, via the utilization of Burlap

## Implementation Details

This was implemented via utilizing KubeJS and modifying the existing create tags and recipes. Adding 1 new item tag, removing 2 recipes and adding 2 new recipes.

## Outcome

Features Removed:
  * Recipe conversion between Sails and Framed Sails has been removed

Features added:
  * 1 new tag ``tfg:usable_in_sail_frame``
    * Added tag to the following items:
      * ``tfc:burlap_cloth``
      * ``tfc:wool_cloth``
  * 2 new recipes
    * Creation of 8 sail frames using 4 lumber and 4 screws
    * Creation of 8 filled sails using 8 sail frames and any item with the tag ``tfg:usable_in_sail_frame``
  
## Additional Information

image
image

Probably none? maybe the fact that you cant convert between sail frames and regular sails


Nebby1999:
I removed the part where i mentioned the creation of colored sails via recipes as that works by just right-clicking a white sail with your dye and said dye is not consumed. So i'm marking this pr as ready for review

Pyritie:
could you add silk cloth to the tag too? since that's the "nylon" you can produce in HV

Nebby1999:
As requested by @Pyritie, i've added silk to the usable in sail frame tag. i appreciate them catching this as i've not advanced the modpack enough to get into electricity shenanigans

PR #731, opened 2025-01-19
Added budding crystals for the other gemstone ores

Pyritie:
## What is the new behavior?
If the ore has a gem associated with it, its new indicator is that gem.
image
image
image
image
image

## Implementation Details
Ore indicators were all adjusted to use these new assets.
I added shimmer entries to them all, so they all glow. If it's too much, I can disable this lol.
Also I updated the field guide for any new indicators.

## Outcome
Better looking indicators

## Additional Information
The new indicators are visible on the minimap. I don't know if that's a problem. They can also be pretty bright at night.

Also I buffed the `deep_sapphire` vein because it was also very rare, and updated the field guide to match.


Pyritie:
I tweaked the textures a bit, so they're less bright.

image
image

Pyritie:
by the way, I tried adding this mod to my game, it works just fine and and it doesn't need any changes so I can't make a PR for it:

https://www.curseforge.com/minecraft/mc-mods/tfc-ruined-world

SverhRazum-Nah:
Are indicators of all ores and gems really necessary? 
I thought indicators were needed for the early stage of the game, when it is difficult to find ore in any other way, and therefore they are available for the necessary ores. But then, a scanner appears, and the indicators are no longer necessary. In addition, these ores are not needed at the beginning of the game anyway, and when they are needed, most likely the scanner will already be ready. 🤔

Pyritie:
> Are indicators of all ores and gems really necessary? I thought indicators were needed for the early stage of the game, when it is difficult to find ore in any other way, and therefore they are available for the necessary ores. But then, a scanner appears, and the indicators are no longer necessary. In addition, these ores are not needed at the beginning of the game anyway, and when they are needed, most likely the scanner will already be ready. 🤔

no, but I think they look kinda cool

PR #732, opened 2025-01-19
Hanging Sign Metal Dupe fix

### LINKED ISSUE #679: [🐛]: Hanging sign recycling dupe

- [X] No, it was not.

### Version modpack used

0.7.15

### Environment

Official TFG Modern server

### New Worlds

Yes

https://mclo.gs/IljmCRR

### Expected Behavior

Not gaining additional resources

### Actual Behavior

To make 3 signs it takes 2 ingots but to after recycling them we get 3 ingots back.
This way for every 2 ingots put in we get 3 ingots out.
Minus some wood and oxygen.

### Steps to Reproduce

1) Make chains of desired resource
2) Make hanging signs
3) Recycle the signs in arc furnace or in macerator
4) Combine piles/nuggets into ingots
5) Enjoy extra resources

### Additional Information

This dupe works with:
Bismuth Bronze
Black Bronze
Bronze
Copper
Wrought Iron
Steel
Black Steel
Blue Steel
Red Steel

AirRice:
## What is the new behavior?
Fixes the TFC Hanging Signs exploit that allowed users to input 2 ingots worth of metal into making them and then getting 3 ingots worth of metal back by using either an Arc Furnace or Macerator. This was due to the hanging signs being made in such a large quantity by cheap ingredients that the lowest possible output granularity (tiny dusts/nuggets) were too large to capture the material put into their manufacture.

Recipe for Macerating black steel acacia hanging signs before this change:

Recipe after this change:

Since adding even more granularity didn't make much sense, I just made the recipe outputs for hanging sign recycling with the Macerator & Arc furnace, which originally gave 1 full nugget's worth (16mB) of metals back, to give the 16mB only 37.5% of the time.
This should roughly equal to, over time, ~6mB per hanging sign, which is equal to the metal input when crafting them
(9mB per chain x2 -> 3 Hanging Signs = 6mB per sign)

## Implementation Details
Although this leaves the output fully to chance and doesn't make sense if you DO get the metal back since, well, it's more than double what you put in to make it in the first place, I think it can be explained as the chains being too small and delicate to be properly salvaged through such a destructive process.

## Outcome
Fixes #679.

## Additional Information

Alternatives considered:
- implement a new item that's of even more granularity than tiny piles/nuggets, 1/16 of an ingot. Obviously this would be overscoping the fix
- making the output not give metals back at all.
- arguing for some other balancing so the most granular amount of metal is always a nugget. Also overscope and not really fit for this sort of quick fix

This recipe is from GTCEU apparently, and no other kubejs script touches it to my knowledge.

### DIFF
--- a/kubejs/server_scripts/gregtech/recipes.js
+++ b/kubejs/server_scripts/gregtech/recipes.js
@@ -1891,4 +1891,40 @@ const registerGTCEURecipes = (event) => {
         .EUt(96)
         .circuit(2)
     // #endregion
+
+    // #region Fix TFC hanging sign metal dupe for Macerator and Arc Furnace
+
+    const SIGN_METALS = [
+		"copper",
+		"bronze",
+		"black_bronze",
+		"bismuth_bronze",
+		"wrought_iron",
+		"steel",
+		"black_steel",
+		"red_steel",
+		"blue_steel"
+	];
+	
+    SIGN_METALS.forEach(metal => {
+        global.TFC_WOOD_TYPES.forEach(wood => {
+            event.remove(`gtceu:macerator/macerate_wood/hanging_sign/${metal}/${wood}`)
+            event.recipes.gtceu.macerator(`gtceu:macerator/macerate_wood/hanging_sign/${metal}/${wood}`)
+                .itemInputs(`tfc:wood/hanging_sign/${metal}/${wood}`)
+                .itemOutputs('gtceu:wood_dust')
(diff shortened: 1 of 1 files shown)


Egogorka:
Wouldn't it be easier to change the amount of crafted signs?

AirRice:
@Egogorka I didn't want to do that as it would nerf the crafting of signs entirely, which doesn't seem fair as it was a perfectly serviceable recipe for a fully decorative block.

This addresses the exploit only.

PR #733, opened 2025-01-19
Fix collapsible brick slab blocks

Egogorka:
## What is the new behavior?
This PR fixes collapsible bricks. Bricks are supposed to be sturdier than raw/smooth rock blocks. But even if bricks are not collapsible, brick slab blocks are for some reason. And because they can be used for decoration, it can bring a lot of pain.

## Outcome
Brick slab blocks don't collapse

Constant `global.TFC_ROCK_SLAB_BLOCK_TYPES` and tags `tfg:rock_slabs`, `tfg:rock_stairs`, `tfg:rock_walls` got changed contents. But search showed no other usage for them besides in the changed file in the code section for collapsible stuff, so it must be okay.

### DIFF
--- a/kubejs/server_scripts/tfc/tags.js
+++ b/kubejs/server_scripts/tfc/tags.js
@@ -192,6 +192,15 @@ const registerTFCItemTags = (event) => {
             event.add(`tfg:rock_walls`, `tfc:rock/${slabType}/${stoneTypeName}_wall`)
         })
     })
+    
+    // Теги для кирпичных ступенек тфк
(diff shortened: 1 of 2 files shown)


AirRice:
It might be worth noting that this modifies recipes such as `tfg:create/shaped/schematic_table`, `minecraft:grindstone`, `railways/track_create_andesite_narrow_from_steel` etc. as they used the `tfc:rock_slabs`, `tfg:rock_walls` tags for definition.

I think the change is still appropriate, and the recipes changed make sense too to me (you wouldn't really be using stone brick slabs for these purposes), but it should be pointed out.

Issue #464, opened 2024-09-06
Seaweed/Kelp variants can't be turned into Dried Seaweed/Kelp


Hexicube:
### Version modpack Used

0.7.12

### Environment

SP

### New Worlds

Yes

### Expected Behavior

Winged/Leafy Kelp can be dried into Dried Kelp and Seaweed (NOT Fresh Seaweed) can be dried into Dried Seaweed.

### Actual Behavior

Only Giant Kelp Flower and Fresh Seaweed work for these recipes.

### Additional Information

This makes acquiring soda ash far more annoying to acquire in bulk if you lack saltpeter as an alternative, especially as seaweed on beaches isn't fresh and has no other purpose.

Not sure if this counts as a feature request instead as nothing is "broken" but it definitely looks unintentional.

PR #742, opened 2025-01-21
Add seaweed and kelp drying recipes

Metriximor:
## What is the new behavior?
Adds drying recipes to seaweed and kelp, just like fresh seaweed and giant kelp flowers have currently.
They are set to 200 temperature, which is the same defined for the recipes in tfc:
https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/1.21.x/src/generated/resources/data/tfc/recipe/heating/food/dried_seaweed.json
And their corresponding item heat capacity is set to 1.0, just like the existing seaweed/kelp recipes.

This is related to this issue: https://github.com/TerraFirmaGreg-Team/Modpack-Modern/issues/464

## Implementation Details
Maybe instead of hardcoding the recipes a tag system could be added to the seaweeds and kelps, but alas this is my first time working with kubejs and I have no idea how to set that up. If there is no big push for it, this should cover most bases.

## Outcome
Recipes to dry kelp and seaweed are added.
Fixes: #464 

We're not using tags, so other seaweed recipes like the plantball are missing. These recipes also turn non-edible items into edible ones, which can be a bit exploity since non-edible items don't rot.

### DIFF
--- a/kubejs/server_scripts/tfc/data.js
+++ b/kubejs/server_scripts/tfc/data.js
@@ -113,6 +113,11 @@ const registerTFCHeats = (event) => {
 
     // Ведро из красной
     event.itemHeat('tfc:metal/bucket/red_steel', 1.429, 924, 1232)
+
+    // Seaweed and Kelp
+    event.itemHeat('tfc:groundcover/seaweed', 1.0, null, null)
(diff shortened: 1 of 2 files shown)


Metriximor:
Nevermind, this isn't working ingame, despite EMI showing the recipe

PR #744, opened 2025-01-22
Changed LV casing/hull recipes (again)

Pyritie:
## What is the new behavior?
LV casings are now always 8 steel plates, both in a crafting grid and in an assembler. Everything LV arc furnaces/macerates down into 8 steel ingots/dust.
LV hulls now require 2 red steel plates instead of 2 wrought iron plates.
MV is unchanged, and then polyethylene can be used to make casings without the red steel.

## Implementation Details
I used the temporary code from 0.7.16 that @/fluffle wrote.

## Outcome
The colored steel is still required, but it's not nearly as annoying as before. Blue steel is popular for tools and lamps so people will want to make a bunch of that anyway.
I liked that people were automating their TFC blast furnace before because of how much steel they needed.

## Additional Information
The all colored steel change in 0.7.17 is very unpopular and some people aren't updating to the latest version because of this change. I think this new change is a fair compromise.

### DIFF
--- a/kubejs/server_scripts/gregtech/recipes.js
+++ b/kubejs/server_scripts/gregtech/recipes.js
@@ -1282,16 +1282,32 @@ const registerGTCEURecipes = (event) => {
 
     //#endregion
 
-    //#region remove LV casing exploit
+    //#region LV casings and hulls
 	
-  event.remove({ id: 'gtceu:assembler/casing_lv' })
+	event.replaceInput('gtceu:shaped/casing_lv', '#forge:plates/red_steel', '#forge:plates/steel')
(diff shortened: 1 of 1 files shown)


SverhRazum-Nah:
> ## What is the new behavior?
> LV casings are now always 8 steel plates, both in a crafting grid and in an assembler. Everything LV arc furnaces/macerates down into 8 steel ingots/dust. LV hulls now require 2 red steel plates instead of 2 wrought iron plates. MV is unchanged, and then polyethylene can be used to make casings without the red steel.

Лучшее решение эвер. А то это ужасно напрягало, что для меня MV корпуса дешевле чем LV, и намного менее геморные 

The best decision ever. It was terribly annoying that for me MV cases are cheaper than LV, and much less pain

PR #745, opened 2025-01-22
Add glowstone as infinite lamp fuel

CaptainGold1:
## What is the new behavior?
Adds glowstone as a TFC lamp fuel, it is usable in all types of lamps and burns forever, like lava. The upside is that it is not restricted to blue steel lamps, so you can use other TFC lamps for aesthetics.

## Implementation Details
Added gtceu:glowstone as a lamp fuel with a consumption rate of -1. This is identical to how lava in blue steel lamps works in the TFC code.

## Outcome
This allows TFC lamps (besides blue steel) to be a viable source of infinite light. They are aesthetically pleasing, so it is nice to keep them as an option. To me, it is balanced to allow it in all lamps, not just colored steels, since lava is infinite and glowstone is far harder to produce, offsetting the lower cost of allowing cheaper lamps such as bronze.

## Additional Information
image

### DIFF
--- a/kubejs/server_scripts/tfc/data.js
+++ b/kubejs/server_scripts/tfc/data.js
@@ -126,6 +126,7 @@ const registerTFCFuels = (event) => {
 
 const registerTFCLampFuels = (event) => {
     event.lampFuel('gtceu:creosote', '#tfc:lamps', 1000)
+    event.lampFuel("gtceu:glowstone", "#tfc:lamps", -1)
 }
 
 const registerTFCFertilizers = (event) => {


Pyritie:
https://www.reddit.com/r/TerraFirmaGreg/comments/1i9t14o/comment/m9rgw4s/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

RubberDuckyDJ:
this should be added to the quest book since there is no way to find out that this is a feature in game as of now.

Pyritie:
it will be in the next version I think... can't remember if I did it in the 0.9 or space branch

PR #753, opened 2025-01-24
Adds heat values for items with valid heating recipes but no heat values

CaptainGold1:
## What is the new behavior?
Fixes the issue where crushed/processed ores do not heat up and thus can never melt into anything in the crucible/blast furnace. Also removed the page in the blast furnace tips that detailed which inputs would work, since all inputs should now work so the page is now irrelevant.

## Implementation Details
Added item heat values through makeItemHeatByTagPrefix() in tfc/data.js for the various stages of ore processing. This should not have any adverse affects, as far as I can tell, as it should only affect ores.

## Outcome
Now you can melt your purified ores directly into liquid, if you so desire. Although it is incredibly stupid to do so, but to each their own.

## Additional Information
image

### DIFF
--- a/kubejs/assets/tfc/patchouli_books/field_guide/en_us/entries/tfg_tips/blast_furnace_tips.json
+++ b/kubejs/assets/tfc/patchouli_books/field_guide/en_us/entries/tfg_tips/blast_furnace_tips.json
@@ -27,12 +27,6 @@
 		},
 		{
 			"type": "patchouli:spotlight",
-			"title": "Valid Inputs",
-			"item": "gtceu:iron_dust",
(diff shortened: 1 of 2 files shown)


CaptainGold1:
Oh wait, blocks don't melt down either, going to add the heat values for those real quick

CaptainGold1:
Okay, now blocks and long rods have heat values too, so they can be melted down as well. I might look into there possibly being some way to overhaul the way heating recipes are currently generated?

CaptainGold1:
Side note, is there any chance that the source code for TFG-Core will be released any time soon? Doing an overhaul of how item heats work so that different metals have different heat capacities would be fairly easy if I could edit TFG-Core to include heat capacities for every metal.

Exzept1on:
I want us to move away from using TFG-Core in a lot of things related to the modpack, as it was done in 0.8.2, almost everything is done there based on KJS, if someone wants to port the code or write it again, why not.

PR #756, opened 2025-01-24
Fixed cassiterite heating "exploit"

CaptainGold1:
## What is the new behavior?
Cassiterite dust and cassiterite sand dust now correctly melt down to 144 mb and 108 mb respectively of tin. All stages of ore processing have also been edited, making all stages correct proportionally to cassiterite and cassiterite sand's correct richness percentage (100% and 75% respectively).

## Implementation Details
A new function was created in tfc/constants to calculate the yield for "processed" ores. If the percent yield plugged in for a "processed" ore is greater than 100, it calculates the correct percent per item by roughly "estimating" how many output items would be produced by macerating the ore and then dividing the percentage by that. This works correctly for cassiterite and cassiterite sand. 
The only heating recipes that now use the new function are the heating recipes for crushed, purified, and refined ores, dusts and their small/tiny versions, and impure dusts. This means that raw ores are not touched.

An alternate solution would be to have the new calcAmountOfMetalProcessed function cap the percent at 100. This has the benefit of not "breaking" for possible future ores that may need to output more than 2 crushed ore per raw ore and simultaneously have a low yield, but has the drawback of making all ores with a percent yield greater than 100 equivalent, i.e. cassiterite and cassiterite sands would have equal yields when melting down.

Another alternate solution would be to make macerating raw cassiterite/cassiterite sand produce only 1 crushed ore, but make smelting their dusts/all other stages of processing return 2 ingots.

## Outcome
Currently, melting cassiterite/cassiterite sand down in a crucible returns more metal than smelting it in a furnace, since the percent yield system did not correctly account for the fact that cassiterite's yield is doubled on the crushing step, not on the smelting step. Now, melting these ores is in line with how it should work. Additionally, if more ores are added with > 100% richness, this solution should work out of the box for them.

## Additional Information
image
image

### DIFF
--- a/kubejs/server_scripts/tfc/recipes.js
+++ b/kubejs/server_scripts/tfc/recipes.js
@@ -1159,7 +1159,7 @@ const registerTFCRecipes = (event) => {
                 if (!tinyDust.isEmpty()) {
                    
                     event.recipes.tfc.heating(tinyDust, tfcProperty.getMeltTemp())
-                        .resultFluid(Fluid.of(outputMaterial.getFluid(), global.calcAmountOfMetal(16, tfcProperty.getPercentOfMaterial())))
+                        .resultFluid(Fluid.of(outputMaterial.getFluid(), global.calcAmountOfMetalProcessed(16, tfcProperty.getPercentOfMaterial())))
                         .id(`tfg:heating/tiny_dust/${material.getName()}`)
                     
                 }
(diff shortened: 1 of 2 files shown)


CaptainGold1:
I am curious what other people think the best solution to this would be, from the 3 solutions I've proposed (the one implemented here and the two alternate solutions), or if you have another, better solution.

PR #758, opened 2025-01-24
Get small dust from failed pressing

Nebby1999:
## What is the new behavior?
This PR changes the behaviour of using the mechanical press to get plates. Instead of having a 80% chance of succeed and when failing returning nothing, when it fails, the player is awarded with 1 small dust of the material. This way press is still a gambling method of getting plates but when it fails at least you get something in return (even if you'd need 4 fails to get another attempt)

## Implementation Details
The pressing recipe of create only accepts a % based chance of multiple outputs, this means that adding a small dust to the array of outputs meant that there was a slight chance for the player to end up with 1 plate and 1 small dust, breaking the laws of conservation of matter

To fix this, the recipes for pressing ingots into plates is now a sequenced assembly that only requires 1 press action to get the result. Which has a 80% chance to be the plate, and if the roll fails, you get a small dust

## Outcome
Creating plates with Create pressing is now slightly less infuriating, still can end with an overall loss of materials.

## Additional Information

New recipe using sequenced assembly, random salvage is only small dusts. Plate retains its 80% chance output.

image

Other than the cursed aura there doesnt seem to be any issues.

### DIFF
--- a/kubejs/server_scripts/gregtech/recipes.js
+++ b/kubejs/server_scripts/gregtech/recipes.js
@@ -1477,7 +1477,6 @@ const registerGTCEURecipes = (event) => {
             let plateStack = ChemicalHelper.get(TagPrefix.plate, material, 1)
             let blockStack = ChemicalHelper.get(TagPrefix.block, material, 1)
             let smallDustStack = ChemicalHelper.get(TagPrefix.dustSmall, material, 1)
-            console.log(smallDustStack);
             
 
             let matAmount = TagPrefix.block.getMaterialAmount(material) / GTValues.M;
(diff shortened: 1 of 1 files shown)


Egogorka:
Now it's even more (although only slightly more) better than forge hammer. Forge hammer has 2/3 ratio, and press now has not 4/5 ratio in long run, but 0.8*1+0.2*1/4*(0.8*1 + 0.2*1/4*...), which gives 0.8/0.95 ~ 84% of getting a plate.

Nebby1999:
I'm not good at math, lol, i think i should fix that ratio but i suck at it.

Pyritie:
it's fine imo, the dust is just a consolation prize

PR #759, opened 2025-01-25
Fixes wool cloth recipe conflict

CaptainGold1:
## What is the new behavior?
Fixes a recipe conflict for wool cloth in the assembler by adding a circuit to it.

## Outcome
Now wool blocks can be crafted instead of wool cloth in the assembler.

## Additional Information
I haven't personally tested the change yet but @KorGgenT did and it works fine for him.


KorGgenT:
image

PR #766, opened 2025-01-27
Remove buggy Pumpkin Chunks recipe

CaitlynMC:
## What is the new behavior?
This recipe for Pumpkin Chunks is easily exploited (gives fresh chunks from rotten pumpkins). It is also unnecessary because TFC already has a working recipe for making Pumpkin Chunks with a hammer

## Implementation Details
It might be possible to create a working recipe using the GT knife that detects rotten pumpkins properly, but it isn't really needed because TFC already has a recipe that works fine

## Outcome
Removes the recipe for making Pumpkin Chunks using the GT knife


Exzept1on:
Bad fix.

Try something like that:
https://github.com/TerraFirmaGreg-Team/Modpack-Modern/blob/9ec15a4254a5ae55c0faada24ca8fb6b39eeb1fd/kubejs/server_scripts/tfc/recipes.js#L3384

Pyritie (approved):

Pyritie:
Changed it to use the correct advanced shapeless crafting recipe type

Nebby1999 (approved):

Issue #444, opened 2024-08-20
Field guide is missing recipes


hownioni:
### Version modpack Used

0.7.12

### Environment

Singleplayer

### New Worlds

Yes

### Expected Behavior

For the recipe images for knapping to appear in the introduction section.

### Actual Behavior

All recipe images for knapping missing from introduction section, except for javelin.

### Steps to Reproduce

Open field guide, go to introduction section and everything will be in blank

### Additional Information

I haven't yet tested to see if this happens in other sections as I'm just starting to play the modpack, but it could be possible that a whole lot of recipe images are missing

DragonKitten:
can confirm, just installed the latest version of the pack and am getting the exact same result

WaterMelwin:
So use jei instead

PR #781, opened 2025-02-04
fixed rock knapping in field guide

GamerDadDave:
## What is the new behavior?
Rock knapping recipes now show up in the introduction section of the field guide.

## Implementation Details
Just needed to change to the tfg rock knapping, which replaced the tfc rock knapping.

## Outcome
Fixes: #444

I was unable to fix the Italian because I couldn't find the TFC guide book in Italian.


Pyritie (approved):

Pyritie:
nebby - they made basically the same changes that I incidentally did in my own fork, so I think it's ok to merge so they get some credit too 😛

Nebby1999 (approved):

PR #788, opened 2025-02-09
reduce recipe requirements for sticky resin in fluid solidifier

KorGgenT:
## What is the new behavior?
changes the recipe for sticky resin in the fluid solidifier to not take sulfur dust.
fluid_solidifier_fluid_solidifier_latex_heating

## Additional Information
by design, this pack is obviously meant to become easier and less of a strain on resources that you dig from the ground as you progress farther in the gregtech tree. not only is the fluid solidifier recipe currently worse than the vat recipe, but it takes an enormous amount of sulfur to utilize latex at all. since tree taps are a pretty cool way of collecting resources, you should be rewarded even in a small way by setting up some kind of automated system - in a lot of ways it is really the first resource that would be more convenient to automate  as kapok don't produce latex at all times.

generally one would eventually move on to some kind of centrifuge setup, but this seems like a reasonable avenue at least for some portion of time in the early game.


Pyritie (approved):

kubejs/server_scripts/gregtech/recipes.js
```diff
@@ -126,6 +126,14 @@ const registerGTCEURecipes = (event) => {
 		.itemOutput('gtceu:sticky_resin')
 		.id('tfg:pot/sticky_resin_from_conifer_pitch')
 
+	event.remove({ id: 'tfg:fluid_solidifier/latex_heating' })
```
Nebby1999: Is there a specific recipe from TFG that has this ID?

If so i think it'd be better to just, remove it from the source instead of removing it here then adding it again later.

Nebby1999 (changes requested):

kubejs/server_scripts/gregtech/recipes.js
Pyritie: I can't remember what this is for 🤔 missed it when merging, maybe it was from back when a bunch of recipes were in TFG-Core?

Pyritie: fixed

Pyritie (approved):

Nebby1999 (approved):

PR #800, opened 2025-02-23
Remove graphite to dust macerator recipe

ZackH01:
## What is the new behavior?
Macerating raw graphite ores will now always give crushed ores. Before it would give dust instead which significantly reduced the yield of graphite from ore processing later in the game.

## Implementation Details
Moved the graphite to dust quern recipes to a new global TFC_QUERN_GRAPHITE_RECIPE_COMPONENTS instead of being with TFC_QUERN_POWDER_RECIPE_COMPONENTS and then added to the script code to add each of these recipes to just the quern and millstone.

## Outcome
Fixed issue where macerating raw graphite ores would give dust instead of crushed ore.

### DIFF
--- a/kubejs/server_scripts/tfc/recipes.js
+++ b/kubejs/server_scripts/tfc/recipes.js
@@ -2587,6 +2587,16 @@ const registerTFCRecipes = (event) => {
 
     //#endregion
 
+    //#region Graphite recipes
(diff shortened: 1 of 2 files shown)


Pyritie:
No longer applicable to 0.9, the quern and millstone and macerator share recipes now

PR #822, opened 2025-03-20
recipes.js

TomPlop:
## What is the new behavior?
Adding the recipes for the new items from ExtendedAE.

## Implementation Details
-Adding the Matrix Multiblock acting as a Pattern Provider/Molecular Crafter 
-Adding the Oversized Interface which is an Extended Interface but each slots have a higher capacity (8 times)

Needs ExtendedAE 1.20-1.3.5 at least

**Please fill in as much useful information as possible. Also, please remove all unused sections, including this and the other explanations.**


TomPlop:
Wrong merging

PR #908, opened 2025-04-22
Update small vessel capacity

oosyrag:
## What is the new behavior?
TFC cost per ingot was 100mB, and the small vessel held 3000mB for up to 30 ingots by default. TFG changed the ingot amount to 144mB, but small vessel capacity never got changed.

## Implementation Details

~~Changed smallVesselCapcity from 3000 to 4608, or 32*144. Because 32 is nicer than 30.~~

Alternative options - 4,320 (30 * 144), 3024 (21 * 144), or 2880 (20 * 144), to divide evenly by 144.

Updated: smallVesselCapacity 3000 to 3024 (divides evenly into 21), crucibleCapacity from 4000 to 4032 (divides into 28), to stay closest to previous values.

### DIFF
--- a/defaultconfigs/tfc-server.toml
+++ b/defaultconfigs/tfc-server.toml
@@ -159,7 +159,7 @@
 		#
 		# Tank capacity of a crucible (in mB).
 		#Range: 0 ~ 2147483645
-		crucibleCapacity = 4000
+		crucibleCapacity = 4032
 		#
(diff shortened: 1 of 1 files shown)


Pyritie:
I like the 20 or 21 number! Feels closer to the original, plus how are you fitting 3 buckets worth of metal into that tiny vessel anyway, let alone 4 😛
Would you be able to change the crucible as well? maybe 21 vs 32? 18 vs 32?

Pyritie (approved):

Redeix (approved):

PR #913, opened 2025-04-25
Added support for armor trims. Added fishing nets. Added more cloth alternatives

Redeix:


kubejs/server_scripts/minecraft/recipes.js
```diff
@@ -972,4 +971,24 @@ const registerMinecraftRecipes = (event) => {
         D: '#forge:storage_blocks/coal',
     }).id('tfg:campfire_coal')
 	//#endregion
+
+    //#region Smithing Table
+    event.recipes.tfc.damage_inputs_shaped_crafting(
```
Pyritie: you don't need to do the "damage inputs" crafting wrappers when you use gt tools, it handles that already for you

kubejs/startup_scripts/gtceu/constants.js
```diff
@@ -183,3 +183,43 @@ global.ORE_BEARING_STONES = [
     'pyroxenite',
     'dripstone'
 ];
+
+global.GTCEU_GEM_TYPES = [
```
Pyritie: This is a big review so it's hard to find, but where is this used? And would it be possible for it to loop over all materials that have the gem flag instead?

Redeix: You're right. This was accidentally left over from development. I can remove it or update it to do what you said. What do you think?

Pyritie: If you're not using it, deleting it sounds good!

Redeix: removed

Pyritie (dismissed):

Redeix:
also allowed vats and pots to be considered oven insulation

Pyritie (dismissed):

Pyritie (approved):

PR #914, opened 2025-04-25
Universal circuits

TomPlop:
Adding the universal circuits to facilitate AE2 automation, shapeless recipes and can't craft themselves to not have issues with AE2

Credit to WithersChat from the Monifactory team for the texture with authorization


kubejs/server_scripts/tfg/universal_circuit.js
Pyritie, line 1: These should be moved into tags.js and recipes-miscellaneous.js

Pyritie (changes requested):

kubejs/server_scripts/tfg/universal_circuit.js
TomPlop: yeah I guess I did it that way before when I was trying to do it with the tag method

Pyritie: If they need to share the const array, put it in startup_scripts/tfg/constants.js and rename it like `globals.UNIVERSAL_CIRCUIT_TIERS` or something

TomPlop: Fixed

Pyritie (approved):

Redeix (approved):

PR #916, opened 2025-04-26
Add Brine/Vinegar to GT Machines and Fix Olive Oil recipe to fit with VAT recipe

TomPlop:
It was two olives when it should be one.


Redeix (approved):

Redeix:
Only issue is that gt recipes still allow you to use rotten ingredients. But I say for now we should just let the recipes be used anyways.

Pyritie (approved):

PR #949, opened 2025-05-04
Rubber processing recipes audit

AirRice:
## What is the new behavior?
This PR seeks to address the fact that latex tapped from trees is no longer the main source of natural rubber. This was very odd and unrealistic IMO so I am making a bunch of recipe changes that hopefully better reflect the IRL method of natural rubber vulcanisation. 

## Implementation Details
The PR leverages the newly added "vulcanized latex" fluid that was introduced in the rubber gloves implementation. 

IRL ancient cultures (such as those from the mesoamericas; the Olmecs were named for their rubbermaking) treated latex with sulfur-containing sap to increase the latex's elasticity and keep it stable. People now use formic acid for a similar purpose to coagulate it for initial processing. This is what we call 'raw rubber pulp' in GTCEu as far as I can tell. Being able to make raw rubber pulp by 'cooking' latex with sulfur therefore made sense before the recipe was changed, but it was changed for a good reason (not making people be softlocked due to weather conditions stopping them from tapping latex). Hence the following changes:

- Latex is now the main precursor to (almost) all rubber
- Latex is obtained either by tapping trees (when the temperature allows) or extracting it using either rubber tree logs/saplings/leaves.
- Latex is first processed into Vulcanized Latex (cook with sulfur powder), and then 'dried' (vacuum chamber, solidifier) into Raw Rubber Pulp.
- Due to this change both Sticky Resin recipes (Conifer pitch and Latex) use wood ash now instead of sulfur powder.
- Preserves the centrifuge Raw Rubber Pulp recipe because it acts as a bit of tech progression.

I didn't fully test the exact values so they might need some further testing. I also may have missed some side paths that allow skipping the entire process, etc.

## Outcome
- Latex can now be extracted from rubber trees' logs, saplings, leaves via the vacuum chamber (ULV) or extractor(LV).
 -1000mB per log, 100mB per sapling, 50mB per leaves; Extractor recipes do this in bulk and output 1000mB per operation.
- Dandelions and Goldenrods can, with added sodium hydroxide, be turned into latex in a chem reactor.
- Sticky resin recipes (from both conifer pitch and latex) require wood ash now instead of sulfur.
- Latex -> Vulcanized latex recipe now requires sulfur instead of wood ash.
- Vulcanized latex now becomes raw rubber pulp in a solidifier or vacuum chamber at a rate of 250mB/raw pulp.

## Additional Information
Also includes Lang string changes for the patchouli book to reflect the changed recipes.

N/A


Pyritie:
Changes to the lang files also need to be made over in [Tools-Modern](https://github.com/TerraFirmaGreg-Team/Tools-Modern/tree/dev/LanguageMerger/LanguageFiles/tfg) so they don't get lost!

kubejs/server_scripts/vintage_improvements/recipes.js
```diff
@@ -570,6 +570,39 @@ function registerVintageImprovementsRecipes(event) {
 		processingTime: 100
 	}).id('tfg:vi/vacuum/rubber')
 
+	// #region Vaccuming rubber wood stuff for latex
+	// This would be a lot nicer if VI allowed for tags as recipe inputs but oh well...
+	Ingredient.of("#tfg:latex_logs").itemIds.forEach(log =>
+		event.custom({
+			type: 'vintageimprovements:vacuumizing',
+			ingredients: [{ item: `${log}` }],
```
Pyritie: You can change this to `ingredients: [{ tag: 'tfg:latex_logs' }]` and then you can remove the foreach

kubejs/server_scripts/vintage_improvements/recipes.js
```diff
@@ -570,6 +570,39 @@ function registerVintageImprovementsRecipes(event) {
...
+			results: [{ fluid: 'tfg:latex', amount: 1000 }],
```
Pyritie: I do worry that a whole 1000mB per log in 15 seconds is going to be significantly faster than bothering with tapping, especially since people will already have a vacuum chamber for progression anyway

kubejs/server_scripts/vintage_improvements/recipes.js
```diff
@@ -570,6 +570,39 @@ function registerVintageImprovementsRecipes(event) {
...
+		}).id(`tfg:vi/vacuumizing/latex_from_${sapling.split('/')[2]}_sapling`)
+	)
+	Ingredient.of("#tfg:rubber_leaves").itemIds.forEach(leaves =>
+		event.custom({
+			type: 'vintageimprovements:vacuumizing',
+			ingredients: [{ item: `${leaves}` }],
+			results: [{ fluid: 'tfg:latex', amount: 50 }],
+			processingTime: 30
+		}).id(`tfg:vi/vacuumizing/latex_from_${leaves.split('/')[2]}_leaves`)
+	)
```
Pyritie: were you going to add one for the rubber plants here too?

AirRice: I was thinking of this, but it would be awkward with trying to keep a certain parity with the extractor recipes - since Goldenrods & Dandelions are both yellow dye plants, they already have an extractor recipe for those. 

Maybe I will add one here since the LV recipe is in a chemical reactor.

Pyritie: ah good point

Pyritie (approved):

Issue #952, opened 2025-05-05
[🪶]: Vessel Limit not Specified Causing Unexpected Results


Zakarovsk:
### Was this feature suggested before?

- [x] No, it was not.

### Version modpack used

0.9.2

### Related problem

It turns out that the **Ceramic Vessel** has an undocumented maximum liquid capacity of 3024 mB. This limit isn't indicated anywhere in the interface, which can lead to unexpected behavior when it is exceeded. The vessel still allows you to add more solid items beyond this threshold, but doing so may result in errors—such as producing an unknown molten liquid instead of the expected alloy. 

For example, I attempted to create Bronze using a 10%-90% Tin-to-Copper ratio by adding 480 mB of Tin and 4240 mB of Copper. The Vessel even predicts the wrong output of 4720 mB of Bronze. Since this combination exceeded the vessel’s capacity, the Tin melted first, but the excess Copper caused an overflow, corrupting the mixture and producing an unknown result.

### Your solution

A practical improvement would be to either limit the amount of solid items (in mB) that can be inserted into the vessel to match the 3024 mB capacity, or—more simply—add a warning in the expected output tooltip when the total projected liquid output exceeds this threshold.

Redeix:
Quick fix for this with #962

PR #962, opened 2025-05-07
Quick Fixes

Redeix:
- Fixed hide duping recipes.
- Added quest book and field guide instances for fishing nets.
- Fixed rebar support recipe crash.
- Fixed red steel flask dupe.
- Fixed red alloy recipe inconsistency.
- Fixed blue steel backpack block lang.


Redeix:
Change for #952. Can probably think of a more elegant solution later.

PR #985, opened 2025-05-11
Add wrought iron alternatives for early game cast iron recipes

ZackH01:
## What is the new behavior?
Added alternative recipes for wooden crates, copper diving boots and the helve hammer that use wrought iron instead of cast iron.

## Implementation Details
Added a new tag that includes cast/wrought iron double ingots (this is used for the helve hammer recipes) and made use of the create deco tag for the copper diving boots.

## Additional Information
This change makes it so that wrought iron can be used as well as cast iron for these early game recipes.

### DIFF
--- a/kubejs/server_scripts/tfg/tags.js
+++ b/kubejs/server_scripts/tfg/tags.js
@@ -129,6 +129,12 @@ const registerTFGItemTags = (event) => {
 
 	// #endregion
 
+	// Use either cast or wrought iron
+	event.add('forge:double_iron_ingots', 'tfc:metal/double_ingot/cast_iron')
+	event.add('forge:double_iron_ingots', 'gtceu:wrought_iron_double_ingot')
+
+	// #endregion
+
 	// #region 0.7.19 -> 0.9 conversion
 
 	event.add('c:hidden_from_recipe_viewers', 'treetap:tap')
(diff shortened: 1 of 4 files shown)


Pyritie:
What's in `#createdeco:internal/ingots/iron_ingots`? is it just the cast iron and wrought iron ingots?

kubejs/server_scripts/tfg/tags.js
```diff
@@ -129,6 +129,12 @@ const registerTFGItemTags = (event) => {
 
 	// #endregion
 
+	// Use either cast or wrought iron
+	event.add('forge:double_iron_ingots', 'tfc:metal/double_ingot/cast_iron')
+	event.add('forge:double_iron_ingots', 'gtceu:wrought_iron_double_ingot')
```
Pyritie: would prefer if this was something like:

`event.add('forge:double_iron_ingots', '#forge:ingots/iron')`
`event.add('forge:double_iron_ingots', '#forge:ingots/wrought_iron')`

Pyritie (changes requested):

kubejs/server_scripts/tfg/tags.js
SpicyNoodle5: Made that change. The #createdeco:internal/ingots/iron_ingots tag is just cast and wrought iron ingots, but I've now changed the diving boots to not use that tag anyway. Probably safer in case that tag ever gets changed at some point.

Pyritie (approved):

kubejs/server_scripts/tfg/tags.js
Pyritie: ah, cool

Redeix (approved):

PR #993, opened 2025-05-13
HV Quest with Moon

TomPlop:
## What is the new behavior?
Adding the new chapter for HV that includes the Moon, also the removal of Eye of Ender without Chemical Bath and fixes on other quests to follow up with HV Chapter.


Pyritie:
looks like this needs to resolve conflicts

TomPlop:
I don’t understand what happened. The branch merging I did with Redeix didn’t work well. I will update everything manually I guess.

Issue #963, opened 2025-05-07
[💥]: JEI Crash when attempting to auto move items to craft with a durability item.


Redeix:
### Was this crash report before?

- [x] No, it was not.

### Version modpack used

0.9.4

### Environment

All

### Side Crashed

Client

### New Worlds

Yes

### Please provide a link to your latest.log, if relevant

https://mclo.gs/VLeX2PL

### Steps to Reproduce

Some items will crash when you hover over the "+" to auto move them to the crafting table. You need to be in the crafting table menu. For example the recipe for wrought iron mortar.

### Additional Information

I am creating this general report for people to post any recipes that create this crash in the comments so we do not get too many duplicate reports.

Redeix:
#958, #926, #805

oosyrag:
Added fixes for iron flask and mortars. I believe the face mask was an unrelated issue with the bone needle that was addressed by Rediex already (also the with the rebar support in #962).  Although from what I can see the rebar support is still a 2 row shaped recipe... I'll test that again shortly. Lets keep this issue open for future reported trouble cases.

Redeix:
I did not change the bone needle recipes

joendter:
I think I have this bug as well for the recipe for the coke oven hatch:
https://mclo.gs/oXJr6XR

Pyritie:
I've removed most of the problematic recipes in 0.10, since EMI doesn't look like they'll get around to fixing this any time soon

PR #1007, opened 2025-05-14
Fix for JEI auto-input crash for mortars and steel flasks

oosyrag:
## What is the new behavior?

Fixes crash from hovering over JEI auto input button for certain recipes - mortars and steel flasks.

## Implementation Details

Per https://github.com/emilyploszaj/emi/issues/942#issuecomment-2865571709, fixed by padding shaped recipes out to include a third row. Thanks @Notenoughmail!

## Outcome

Addresses https://github.com/TerraFirmaGreg-Team/Modpack-Modern/issues/963, although maybe that issue should be kept open for additional items/cases? This is a bandaid fix for specific items that were reported, there may be more out there.


kubejs/server_scripts/tfg/recipes.js
```diff
@@ -14,4 +14,29 @@ const registerTFGRecipes = (event) => {
 	registerTFGSupportRecipes(event)
 	registerTFGFishingNetsRecipes(event)
 	registerTFGConvertingRecipes(event)
+
+	//Fixes for 2x2 crafts crashing when auto inputting to crafting grid via JEI by padding recipes to 3x3 with empty space.
+	//https://github.com/emilyploszaj/emi/issues/942
+
+	event.shaped('waterflasks:iron_flask', [
```
Pyritie: These should be in water_flasks/recipes.js instead!

Pyritie (changes requested):

kubejs/server_scripts/tfg/recipes.js
Pyritie: (also maybe we could change them to `#forge:cloth` instead?)

Notenoughmail:
have you verified this fixes the crash when using the crafting upgrade in Sophisticated Backpacks? In my testing the game never crashed when using a normal crafting table. I ask because, as far as I'm aware, empty pattern rows are simply discarded by the deserializer and thus won't change EMI's behavior.

Also, shouldn't the water flask recipes be wrapped in TFC's damage inputs recipe type?

oosyrag:
Yep, actually only tested in backpack but didn't try the crafting table. I'll go do that now.

I know previously when first encountering the crash I had confirmed crashing when hovering the + button for both the crafting table and backpack, but it worked fine in the inventory 2x2 grid.

oosyrag:
I don't think it has to do with the shaped recipe size anymore, it has to do with the tfc damage inputs recipe type. When I added that wrapper, it started crashing again. The original mortar recipes also had it and crashed. When I commented out the wrapper, it was ok - same for the mortar. I reduced the iron flask recipe to two rows and it was still okay as well.

I did confirm there's no difference between backpack and crafting bench though - Both are either ok or crash in the same situations.

Is my syntax correct here? The top one crashes, and the bottom one is fine
```
	event.recipes.tfc.damage_inputs_shaped_crafting(
		event.shaped('waterflasks:iron_flask', [
			'AB',
			'CD'
		], {
			A: 'waterflasks:iron_flask',
			B: 'waterflasks:bladder',
			C: '#forge:cloth',
			D: '#forge:tools/knives'
		})
	).id('waterflasks:crafting/repair_iron');

//	event.recipes.tfc.damage_inputs_shaped_crafting(
		event.shaped('waterflasks:iron_flask', [
			'AB',
			'CD'
		], {
			A: 'waterflasks:broken_iron_flask',
			B: 'waterflasks:bladder',
			C: '#forge:cloth',
			D: '#forge:tools/knives'
		})
//	)
	.id('waterflasks:crafting/repair_broken_iron');```

**Comment by Notenoughmail** [comment]
huh, strange. when I was investigating (freshly installed TFG 0.9.5), the game only crashed within the Sophisticated Backpacks crafting, inventory and regular crafting worked regardless of the wrapper recipe type

but, yes the issue is with the damage ingredients wrapper, but also with the number of rows in the shaped recipe's pattern

EMI would crash with
```js
event.recipes.tfc.damage_inputs_shaped_crafting(
  event.shaped('minecraft:dirt', [
    'AB',
    'DD'
  ], {
    A: '#forge:tools/knives',
    B: 'minecraft:obsidian',
    D: 'minecraft:gravel'
  })
);
```
but not with
```js
event.recipes.tfc.damage_inputs_shaped_crafting(
  event.shaped('minecraft:dirt', [
    'AB',
    'D ',
    'D '
  ], {
    A: '#forge:tools/knives',
    B: 'minecraft:obsidian',
    D: 'minecraft:gravel'
  })
)
```
due to an error in its non-vanilla crafting recipe interpreter, so yes your syntax is correct

vanilla crafting recipes should work fine, though you may have to use kube's damage ingredient method to use durability, turning the recipe into one of kube's crafting type, which may also have this problem

all that said, if it doesn't crash and the ingredients are damaged like expected, I suppose its fine as is

oosyrag:
I was unable to get this to work satisfactorily. The latest commit has damage_input wrapped recipes and padded recipes to 3x3, still crashes for me for both the crafting bench and backpack when mouse-over autofill button in JEI. Inventory crafting is fine.

Issue #989, opened 2025-05-12
[💥]: Game crash when trying to open AE2 Guide book


Aerisan:
### Was this crash report before?

- [x] No, it was not.

### Version modpack used

0.9.5

### Environment

Open to LAN

### Side Crashed

Client

### New Worlds

Yes

### Please provide a link to your latest.log, if relevant

https://mclo.gs/xFILUHq

### Steps to Reproduce

Tried to open the '?' help book thing for AE2 from a crafting terminal.
Swapped to "Saving World" screen, then promptly closed.

### Additional Information

I have no idea. Good luck :)

Pyritie:
https://github.com/AppliedEnergistics/GuideME/issues/57

Pyritie:
This is partially fixed in 0.9.6, though to _fully_ fix it we'd need to remove Starlight (a lighting engine optimization mod) and I don't know what side effects that might have. I'll leave the ball in ae2's court for now

PR #1020, opened 2025-05-15
Release: 0.9.6

Pyritie:
🦀 SPRINKLERS ARE FINALLY FINISHED 🦀

GuideME is a new dependency for AE2. Looks like they moved their ingame guide into its own mod, like what create did with its ponder system.


Xikaro:
Why do you need a guideme mod? What's the point of another instruction?

Pyritie:
It's a new required dependency for AE2. You can hover over an AE2 item and hold G I think? to open up a guide, like you can do with Create's ponder system. Without GuideME installed, trying to open the guide on an item will just crash your game (#989)

Xikaro:
How tired I am of them with their manuals and their rigid dependence

Xikaro:
https://legacy.curseforge.com/minecraft/mc-mods/starlightxae

Need to check

Pyritie:
> As the author of Starlight said, it isn't necessary to use Starlight in 1.20.x because Mojang has rewriten the lighting engine and its perfermance is almost the same as Starlight installed.

oh good, removing starlight sounds easy enough to do then

Xikaro:
Have you tried simply deleting the mod?

Xikaro:
Starlight

Pyritie:
Yes, TFG worked fine when I ran it on my own PC, and I didn't get any crashes with the ae2 guide any more

Xikaro:
Let's delete it then
Or use what I threw above

Pyritie:
yeah should be fine to delete it then 👍 (I'm at work at the moment though)

PR #1021, opened 2025-05-16
Bedrock miner for the moon

TomPlop:
-Recipes for each Bedrock Miner (at HV, IV and ZPM)
-Bedrock veins
-Quest at HV for Bedrock Miner

### DIFF
--- a/kubejs/server_scripts/gregtech/data.js
+++ b/kubejs/server_scripts/gregtech/data.js
@@ -73,7 +73,7 @@ const registerGTCEUBedrockOreVeins = (event) => {
     })
 
         // Mica - Talc Silicon Aluminium Potassium Fluorine Caesium
-        event.add('tfg:moon_asbestos', vein => {
+        event.add('tfg:moon_mica', vein => {
         vein.weight(500)
             .size(1)
             .yield(1, 3)
--- a/kubejs/server_scripts/gregtech/data.js
+++ b/kubejs/server_scripts/gregtech/data.js
@@ -20,37 +20,56 @@ const registerGTCEUHeats = (event) => {
 
 const registerGTCEUBedrockOreVeins = (event) => {
 
-    event.add('tfg:bedrock_olivine_vein', vein => {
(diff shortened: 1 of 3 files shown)


kubejs/server_scripts/gregtech/data.js
```diff
@@ -20,37 +20,56 @@ const registerGTCEUHeats = (event) => {
...
-        
-        vein.dimensions('ad_astra:moon')
-        vein.material(GTMaterials.Olivine, 100)
-        vein.weight(1000)
-        vein.yield(30)
-        vein.depletedYield(1)
-        vein.depletionChance(2)
-        vein.depletionAmount(1)
+    // Tin
+        event.add('kubejs:moon_tin', vein => {
```
Pyritie: use tfg namespace instead of kubejs pls

Pyritie (changes requested):

Pyritie:
it'd be cool if there was a bedrock vein for aluminium as well!

TomPlop:
I thought about other veins, I was already a bit reluctant for gold but for Aluminium the issue is it’s a source of Rutile. And there is already going to be bigger veins of Bauxite/Alu/Ilmenite on the moon to mine.

And Rutile is especially important as it’s required for Fusion like Arsenic so I’m thinking we should sit on it for now and wait Mars.

Pyritie:
True, but what about some other mineral that has aluminium? Do any of the other new moon minerals have it? I forget
Otherwise something like lazurite might be good too, iirc its byproducts don't have anything unique

kubejs/server_scripts/gregtech/data.js
TomPlop: Fixed

TomPlop:
Mica vein added with rarity a bit lower than the copper/tin one so nothing too bad even if it can take some time.

Pyritie (approved):

Redeix (approved):

PR #1032, opened 2025-05-17
Fix Electrolyzing lang formate for moon update

TomPlop:


Pyritie (approved):

Pyritie:
if you've changed something in dev, you don't need to do a PR for the space branch too, it'll get pulled in later

TomPlop:
Yeah it's because there is a small difference, for space the quest will have 2 pages and right now it has only one. I could have fit everything on one page and just change the language file but well I don't know it was big.

Pyritie:
ahh ok

PR #1041, opened 2025-05-18
Add extra small gear anvil recipes

SpicyNoodle5:
## What is the new behavior?
Added anvil recipes for red alloy and black steel small gears. This makes these metals consistent with the other TFC metals that have small gear anvil recipes and makes red alloy gears (used in speed controllers) cheaper.

## Implementation Details
Reused code for other small gear anvil recipes to add these cases.


Pyritie (approved):

Pyritie:
removed black steel gear in local, edited PR to the cast iron gear instead

PR #1080, opened 2025-05-26
Reduce rarity for graphite veins

oosyrag:
## What is the new behavior?
Decreases overworld graphite rarity, so that finding graphite with a prospectors pick is more feasible.

## Implementation Details
Change graphite vein rarity from 80 to 60. Previously it was pretty common to not see any graphite at all within range of the big prospector.

## Outcome
Graphite more common, but there are still some areas where its possible to not have any within range of prospector. 
Screenshots of new world at rarity 60:

## Additional Information
Not sure what's going on with my VSC autoformatting... hope that's ok.

### DIFF
--- a/kubejs/data/tfg/worldgen/configured_feature/earth/vein/normal_graphite.json
+++ b/kubejs/data/tfg/worldgen/configured_feature/earth/vein/normal_graphite.json
@@ -4,391 +4,170 @@
   "config": {
     "height": 6,
     "size": 16,
-    "rarity": 80,
+    "rarity": 60,
     "density": 0.4,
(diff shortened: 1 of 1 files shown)


Pyritie:
The vein [needs to be updated](https://github.com/TerraFirmaGreg-Team/Tools-Modern/blob/dev/OresToFieldGuide/data/veins/earth/normal_graphite.json) in Tools-Modern too, then run the tool to generate the file in Modpack-Modern (which will also update the field guide)

oosyrag:
The root problem was something else, closing.

PR #1081, opened 2025-05-26
Harmonize jars size

Zeropol:
## What is the new behavior?
Full jars are the same size as empty jars.
The only effect I see is when placing them with "v", full jars are consistent with empty jars. (take 1/4 of a block, not a full block)

## Implementation Details
Added some  `event.itemSize()` in `\kubejs\server_scripts\tfc\data.js`

Alternative I though :
- changing only the tfc's jars in TFC's folder, and changing the firmalife one's in firmalife folder. Firmalife had no data.js
- changing the weight from *Heavy* to *Medium*.

I put this PR as a draft because I do not understand the last parameter of the `itemSize()` function.
If someone can tell me...
And also because I would want to figure out how to stack more than 4 full jars in shelfs.

## Outcome
Full jars are the same size as empty jars.

## Additional Information

It don't change how many full jars we can put in shelfs (only 4).
It don't change how many jars we place with right click (4).
It don't change any weight because I dont want to break balance.

Before : 
image
After : 
image

I tried this pull request after this discord thread : [Item Size and Weight Deconfliction](https://discord.com/channels/400913133620822016/1376606408199372960)

This error show up about a file I did not touch `Failed to create recipe for type 'tfc:sewing': ItemStack 'result' can't be empty!`
image

Because it don't seem related by the items touched nor the files, I felt free to PR anyway.

Zeropol

### DIFF
--- a/kubejs/server_scripts/tfc/data.js
+++ b/kubejs/server_scripts/tfc/data.js
@@ -209,6 +209,9 @@ const registerTFCItemSize = (event) => {
     event.itemSize('tfg:fishing_net/magnalium', 'large', 'medium', 'magnalium_fishing_net')
 
     event.itemSize('#tfc:ore_pieces', 'very_small', 'very_light', 'tfc_ores')
+	
+    event.itemSize('#tfc:foods/sealed_preserves', 'tiny', 'heavy')
+    event.itemSize('#tfc:foods/preserves', 'tiny', 'heavy')
 
 }
 //#endregion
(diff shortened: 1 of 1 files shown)


Zeropol:
So I suspect a race condition somewhere : 
Sometimes when I load the save, some jars still keep their original size and weight. ( say, only unsealed food jars )
It is not consistent, so after another reload, maybe these unsealed food jars will be fine, but the firmalife one won't.

I need help to sort this out please.

Screenshot : 
image

Pyritie (approved):

Redeix (approved):

PR #1082, opened 2025-05-28
Vulcanized latex using Vacuum Chamber

FernandimModelador:
## What is the new behavior?

adds a recipe to craft vulcanized latex using the vacuum chamber

## Implementation Details

the recipe needs to be heated so its similar as if using a vat to craft vulcanized latex

## Outcome

Now its possible to use the vacuum chamber to craft vulcanized latex

One issue or feature is that when it finishes making vulcanized latex it also starts making raw rubber pulp
if there are still some sulfur powder in the basin

discord
fernandim_modelador

### DIFF
--- a/kubejs/server_scripts/vintage_improvements/recipes.js
+++ b/kubejs/server_scripts/vintage_improvements/recipes.js
@@ -601,6 +601,14 @@ function registerVintageImprovementsRecipes(event) {
 		processingTime: 40
 	}).id('tfg:vi/vacuumizing/latex_from_rubber_plants')
 
+	event.custom({
+		type: 'vintageimprovements:vacuumizing',
+		ingredients: [{ item: 'tfc:powder/sulfur' }, { fluid: 'tfg:latex', amount: 1000 }],
+		results: [{ fluid: 'tfg:vulcanized_latex', amount: 1000 }],
+		heatRequirement: "heated",
+		processingTime: 50
+	}).id('tfg:vi/vacuumizing/vulcanized_latex_from_sulfur')
+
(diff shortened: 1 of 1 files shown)


Pyritie:
@AirRice how do you feel about this, balance-wise?

FernandimModelador:
my reason for this change is that when you are at ulv making your preparations to start making  things for lv
you still need to use either a pot or the vat to make vulcanized latex

Pyritie:
yeah, but the vat can also do it in bulk and it's not that difficult to just batch craft a drum of it at a time - I don't want to make the vat totally useless. In your pull request I would definitely want the time increased to be similar to the vat time if possible (and maybe change it to pressurizer instead of vacuumizer?)

Pyritie:
We had a little discussion about it earlier today, https://discord.com/channels/400913133620822016/1331744604701069393/1377677161128792208

FernandimModelador:
Yeah I kinda of agree with most of the things that were said. My reason to make this pr was that while playing I found it strange that I could do most of the recipes from the vat using a mixer or the vacuum chamber but vulcanized latex wasn't one of them.

So I decided to make one

Pyritie:
That's fair, but also you only need the latex right before you hit LV where you unlock better options, whereas for some of the other vacuum chamber recipes, they're for stuff that you could want a lot earlier, like glue. "Thing is difficult at first but easier once you use it to build a new thing" is the general design philosophy we're sticking to

Pyritie:
Yeah, I wish the vat lid could be automated with redstone or something like you can with barrels, but firmalife is very resistant to any sort of automation 😔

PR #1128, opened 2025-06-05
Fix Scaffolding Frame conflict

JeanRdSz:
## What is the new behavior?
This adds a circuit for the Scaffolding Frame recipe

## Outcome
This will fix the conflict between Scaffolding Frame and Painting recipes

image


TomPlop:
I thought this one had been fix before I don't know why we have an assembler recipe for this one now, it doesn't feel needed.

Redeix:
Yeah, honestly just remove the assembler recipe. There isn't any point in it existing

Issue #1140, opened 2025-06-08
[🐛]: Jar and Wine glass has the same craft in alloy smelter.


BebraNoska:
### Was this bug reported before?

- [x] No, it was not.

### Version modpack used

0.9.10

### Environment

singleplayer

### New Worlds

Unsure

### Please provide a link to your latest.log, if relevant.

https://mclo.gs/J7EjcgQ

### Expected Behavior

expected to craft jars in (steam) alloy smelter

### Actual Behavior

![Image](https://github.com/user-attachments/assets/2070c1c4-36d2-4c80-b71a-57fabd29e53a)
Instead of crafting jars - it crafting wine glasses.

### Steps to Reproduce

1. put a glass batches and a cylinder mold into the alloy smelter
2. wait until it melts
3. get wine glass instead of jar

### Additional Information

btw no matter how you place mold - it still crafts wine glasses

JeanRdSz:
Oopsies. I'll fix that. Thanks for reporting

TomPlop:
Is it fix? Because I'm making a run on fixing small kubejs issues.

PR #1141, opened 2025-06-08
Conflict issues fix

JeanRdSz:
## What is the new behavior?
This will remove the Scaffolding conflicting recipe with Paintings and fix the Wine Glass conflicting with Jar

## Outcome
This will fix the issue #1140 

Kinda hard to find a workaround, Wine Bottles were conflicting with 2 Jar recipes (Hematitic and Silica glass batch), so I changed the Jar recipe instead. Scaffolding frame recipe in the Assembler is just unnecessary, so I removed it.

### DIFF
--- a/kubejs/server_scripts/tfc/recipes.machines.js
+++ b/kubejs/server_scripts/tfc/recipes.machines.js
@@ -355,7 +355,7 @@ function registerTFCMachineRecipes(event) {
 	// Empty Jar
 	event.recipes.gtceu.alloy_smelter('tfc:jar_alloying')
 		.itemInputs('#tfc:glass_batches_tier_2')
-		.notConsumable('gtceu:cylinder_casting_mold')
+		.notConsumable('gtceu:ball_casting_mold')
 		.itemOutputs('tfc:empty_jar')
 		.duration(100)
 		.EUt(2)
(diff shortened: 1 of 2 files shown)


kubejs/server_scripts/tfc/recipes.machines.js
```diff
@@ -355,7 +355,7 @@ function registerTFCMachineRecipes(event) {
 	// Empty Jar
 	event.recipes.gtceu.alloy_smelter('tfc:jar_alloying')
 		.itemInputs('#tfc:glass_batches_tier_2')
-		.notConsumable('gtceu:cylinder_casting_mold')
+		.notConsumable('gtceu:ball_casting_mold')
```
Pyritie, line 358: isn't this one used for glass tubes? for vacuum tubes

Pyritie (changes requested):

kubejs/server_scripts/tfc/recipes.machines.js
TomPlop: it uses glass dust

Pyritie: oh

Pyritie (approved):

Redeix (approved):

PR #1153, opened 2025-06-09
Migmatite conflict fix

JeanRdSz:
## What is the new behavior?
Polished rock recipes will use a glass tube instead of lens in the Laser Engraver machine

## Outcome
This will fix the Polished-Cut Migmatite conflict. Tbh I can hardly call this a good solution, but with Stonecutter coming back, this might even be meaningless
image


Pyritie:
lol the glass tube is funny
imo I would've just added a stonecutting recipe from the raw to the polished and removed the chisel/laser engraver one

JeanRdSz:
Hopefully, it's all good now

Pyritie (approved):

PR #1166, opened 2025-06-15
Add some more food recipes

ofoxsmith:


gustovafing:
Hmmm I though I had already updated the branch earlier

PR #1189, opened 2025-06-19
Nutrition Quest Update

thedarklord6609:
## What is the new behavior?
Adds a small tip about stacking food to combine expiration dates

## Implementation Details
Updates quest description

## Outcome
Eases food storage and the annoyance of having a max stack of 5 food when doing something such as using the grill to make food

## Additional Information
image

Discord: thedarklord6609


Pyritie:
The text is good, but this is the wrong place to put it -- add it to the file over [here](https://github.com/TerraFirmaGreg-Team/Tools-Modern/blob/dev/LanguageMerger/LanguageFiles/tfg/en_us/Quests/stone_age.json) instead!

Issue #1169, opened 2025-06-16
[🐛]: Lunch box item update issue


LARLIOTHI:
### Was this bug reported before?

- [x] No, it was not.

### Version modpack used

9.0.11

### Environment

Essential Multiplayer

### New Worlds

No

### Please provide a link to your latest.log, if relevant.

not so

### Expected Behavior

The lunch box would update anytime an item is moved in it's inventory and the item is moved while it's open.

### Actual Behavior

The lunch box's inventory updates only after you close the inventory.

### Steps to Reproduce

1. Put food in a lunch box
2. close the lunch box
3. open the lunch box back up
4. take the food out of the lunch box
5. click on the lunch box item while it is open
6. Voila, the items have been successfully moved from the lunch box while not updating the inventory of said item

Pyritie:
@Thomasx0

PR #1190, opened 2025-06-19
lunchbox update

Thomasx0:
lunchbox update into Pakku
Refactored KubeJS recipes for GT compatibility


Pyritie:
If the gregtech cells are all compatible, do you still think we need your lunchbox capsule?

Thomasx0:
You're right - we should remove it, Greg's got plenty of his own. Should I rework the PR or will you handle it?

Pyritie:
I can do it

Thomasx0:
My bad — I missed [that](https://github.com/TerraFirmaGreg-Team/Modpack-Modern/issues/1169)

PR #1233, opened 2025-06-30
removed forgiving world, added simply stacked dimensions, added configs

Pyritie:
It doesn't get stuck like forgiving world does, at the expense of teleporting you very literally into the same X/Z coordinates, including inside rock.

The cloud blocks dissipate over time. The nether -> overworld height is just over the main lava level, so it should give you enough time to scramle and put down some bricks before the clouds disappear.


Redeix (approved):

Redeix:
It would be kinda funny if they didn't bring any building blocks

Pyritie:
that's what we call a "skill issue"

PR #1254, opened 2025-07-07
Moon rabbit meat stuff

Redeix:
I reverted the change to use #tfg:foil_packs back to tfg:foil_pack because I want players to have to go through the full recycling process to show off the downsides of single-use packaging. They are very cheap anyways.

I also nerfed nutrition on calorie paste. They should be the best for pure hunger and saturation, but I would rather players make meal bags for nutrition points.


Pyritie:
ah, I missed the part where they could be recycled like that, my bad

PR #1275, opened 2025-07-11
Adding the Large Solar Panel

TomPlop:
-Add tooltip to Ad Astra Solar Panel
-Adding three chemical fluids for LSP
-Fully looped the crafting of Solar Coolant just need to add Helium 3 and Water
-Adding the Large Solar Panel multiblock and recipes
-Adding recipes for new casings for LSP
-Adding Gear flags to dash and rocket alloy tier 1
-Updating the Ad Astra config file to allow 256 rf output for Ad Astra solar panel


kubejs/client_scripts/tooltips.js
Redeix, line 1: Do we know if it's possible to add a lang string here? I haven't tried it myself but it might be possible

TomPlop: You are right I could try later if it works

Pyritie: I've been curious if tooltips can use lang strings for a while now - kubejs, unsurprisingly, doesn't document this

Pyritie (approved):

kubejs/client_scripts/tooltips.js
Redeix: There's a lot it surprisingly doesn't document... We can merge it for now and see if we can figure out how to replace all the tooltips later

PR #1280, opened 2025-07-12
Feature/ulv ev guns

BlueBoat29:
Lots of guns and helper recipes

no lang is defined yet, will do it in the tools modern


pakku-lock.json
```diff
@@ -6826,6 +6826,7 @@
             "id": {
                 "curseforge": "585406"
             },
+            "export": false,
```
Pyritie: revert this line

Redeix (approved):

Issue #1315, opened 2025-07-14
[🐛]: No electric oven recipe for dried kelp


TreyR9:
### Was this bug reported before?

- [x] No, it was not.

### Version modpack used

0.10.0

### Environment

Singleplayer, Forge Server

### New Worlds

Yes

### Please provide a link to your latest.log, if relevant.

N/A

### Expected Behavior

A recipe for dried kelp in the electric oven would be nice

### Actual Behavior

But there is not one :(

### Steps to Reproduce

Step 1: need dried kelp for film
Step 2: put it in oven
Step 3: no dried kelp :(

PR #1319, opened 2025-07-14
Bug fix/Adding in some missing recipes

BlueBoat29:
added kelp and seaweed drying to electric oven (#1315)
added tfc properties to lead
added lead nugget crafting in anvil
added 45-70 crafting in lv assembler
fixed afc wood stomping barrel recipes


Redeix (approved):

Pyritie:
We'll need to also add a texture for lead in the casting channels and for placed ingots, I believe?

Issue #1318, opened 2025-07-14
[🐛]:  slight wording error in HV "Nichrome Ingots" quest


Hee-roe:
### Was this bug reported before?

- [x] No, it was not.

### Version modpack used

0.10.0

### Environment

Singleplayer

### New Worlds

Yes

### Please provide a link to your latest.log, if relevant.

[Not relevant]

### Expected Behavior

"This time, you'll notice that you can't use a Chemical Washer. Instead, you're required to use a Vacuum Freezer."

### Actual Behavior

"This time, you'll notice that you can't use a Chemical Washer. Instead, you're required to use a 'Vacuum Chamber.'"

<img width="1932" height="1900" alt="Image" src="https://github.com/user-attachments/assets/f49f5a34-df08-42ea-a188-87ccbfded25b" />

### Steps to Reproduce

Open quests, go to HV quest tab. "Nichrome Ingots" quest is top right, dependent on the "Ice Cold" Vacuum Freezer quest.

TomPlop:
Hmm weird I will fix it

PR #1333, opened 2025-07-15
Update quests alpha 0 10

### LINKED ISSUE #1291: [🐛]: quest description isnt acurate

- [x] No, it was not.

### Version modpack used

0.10.0

### Environment

singleplayer

### New Worlds

Yes

not revelant

### Expected Behavior

Image

### Actual Behavior

Image

### Steps to Reproduce

...

### Additional Information

Minor problem, not even bug. i guess someone didnt update descryption

TomPlop:
Fix #1291 and #1318
Finished HV Chapter
Begins AE2
Small fixes around

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 - Fixed rotten cheese in moon quest reward @Pyritie
 - Fixed create + greate configs (#1300) @HiddenOwlllll1 @Pyritie
 - Fixed blue steel diving suit looking like netherite (#1293) @Pyritie
+- Fixed anthracite not working with blaze burners @Pyritie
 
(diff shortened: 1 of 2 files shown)


config/ftbquests/quests/chapters/gregtech_energy.snbt
Pyritie, line 1: I moved the "full" version of this quest to space survival, since that's what you'll probably be reading to craft the thing, and left just the controller in the energy chapter - what do you think? I don't think we need the full list in both places

TomPlop: I made it a linked quest so it's actually the same quest now.

PR #1337, opened 2025-07-15
Updated Primitive Creatures

Pyritie:
- Updated primitive creatures to latest
- Comes with some balance changes to the kaolin klayze and graphite glayze
- Also comes with all the illagers. I made loot tables and renamed things and did recipes and all that. Let me know what you think about the loot tables!


Pyritie:
ayyy I got #1337 😎

Redeix (approved):

Redeix:
man what the hell are those names T.T loot seems good but idk what grh is or what the totems do

Pyritie:
"grh" is a "battered wool" item, sort of like a chunk of their primitive clothing. I added a recipe to turn two of it into raw wool.

These totems don't do anything, but you can uncraft them back into clay, sticks, or daub - they're just for flavour from what the author told me.

PR #1378, opened 2025-07-19
Pink Thorium

Nebby1999:
## What is the new behavior?

* Makes thorium pink

## Implementation Details

* I used a colorpicker

## Outcome

* Thorium is now Pink
image

## Additional Information

* I made it pink to turn it into an Obscure Mindustry reference, plus, I saw some developers where looking to recolor it to differentiate from another material.

image

If you don't like pink then this may be a downgrade


Pyritie:
we love pink!!!

Pyritie (approved):

PR #1379, opened 2025-07-19
Steam Fluid Blaze Burner

Nyonyix:
## What is the new behavior?
_This section describes what this PR is about. It should be a clear and concise description concerning what this PR is for, why this PR is needed, and why it should be accepted._
_Linking an issue can be used alternatively to writing a description._

Allow for steam to be used in the fluid blaze burner.

## Implementation Details
_Any implementations in this PR that should be carefully looked over, or that could/should have alternate solutions proposed._

Added steam recipe to fluid blaze burner.

## Outcome
_A short description of what this PR added/fixed/changed/removed._
_For correct linking of issues please use any of the Closes/Fixes/Resolves keywords. Example: When a PR is fixing a bug use "Fixes: #number-of-bug"_

Fluid blaze burner uses 256 steam every ~16 seconds. Subject to change, it's based on my own feel for balance.

Also, Unabe to move the create steam engine over to KJS. Seems to be hard coded to accept water.


Redeix:
I don't really think this should be included. Blaze burners burn fuel, burning steam makes no sense. I already don't like letting lava be used.

Pyritie:
> I don't really think this should be included. Blaze burners burn fuel, burning steam makes no sense. I already don't like letting lava be used.

You already used other fuel to turn the steam into lava, this is just more of a compatibility thing -- there's already a tfg-core mixin to allow steam to be pumped into the tank, but it doesn't work without the blaze burners being heated as well

Pyritie:
If you were able to pump steam into the tank and have it power the steam engines _without_ the heating requirement? that would be perfect - this just feels like an easier workaround, lol

TomPlop:
This is a machine that makes Steam and we input steam into it? I don’t understand at all. This machine takes water and the same fuels as the boiler. What’s the point of this workaround. 

If we follow this design we should be able to input SU into the GT boilers to output steam…

Pyritie:
I think of it more like a machine we input steam into, and it turns that into SU

TomPlop:
That’s a Steam turbine and an alternator? You already input water in the Steam Engine. 

The point of the workaround sounds like it’s about skipping alternator. Why?

Pyritie:
No, it's a way of taking the steam output from your boilers and turning it into SU via the steam engine. If you then connect the SU output into alternators then yeah it's just a wonky workaround for the steam turbines lol

Redeix:
I get it now. Feels very janky tho

Pyritie:
out of curiosity, how did you test the balance of this? 256 steam _feels_ really low, but I don't know

Nyonyix:
> out of curiosity, how did you test the balance of this? 256 steam _feels_ really low, but I don't know

It's based on the solar boiler but also a round number. 256 and 16. Though I'm not familiar with the math behind the scenes so if people have a better value feel free to use it.

Pyritie:
it's a round number in computer terms, sure, but in TFG terms, 1000mb is a round number but also so is 144mb (16*9, common amount mods use for a lot of things since it gives both nuggets and other even dividing amounts)

how about 250mb?

Pyritie (approved):

Redeix:
I will want to integrate this better in the future

Redeix (approved):

PR #1391, opened 2025-07-19
Fixes bug where milling #tfg:hardwood would result in softwood pulp instead of the expected hardwood pulp.

Forsuin:
## What is the new behavior?
This fixes a bug where milling a #tfg:hardwood item would have the resulting product be Softwood Pulp instead of Hardwood Pulp. 

## Implementation Details
I removed the 'gtceu:macerator/macerate_logs' recipe and added a macerator recipe for #tfg:softwood that is the same as the existing 'macerator_hardwood' recipe

## Outcome
Macerating a #tfg:hardwood item results in Hardwood Pulp as expected

## Additional Information

Softwood pulp milling recipe still shows minecraft:logs as input item but I'm not familiar enough with kubejs to know how to fix it. 
image
image

Hardwood Pulp is the correct resulting item though

image

**Discord Username: forsuin**

### DIFF
--- a/kubejs/server_scripts/tfg/recipes.paper.js
+++ b/kubejs/server_scripts/tfg/recipes.paper.js
@@ -46,10 +46,6 @@ function registerTFGPapermakingRecipes(event) {
 		.duration(160)
 		.EUt(7)
 
-	//Replace macerate logs into macerate softwood logs
-	// event.replaceInput({ id: 'gtceu:macerator/macerate_logs' }, '#minecraft:logs', '#tfg:softwood')
-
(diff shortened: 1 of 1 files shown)


Pyritie:
Use `removeMaceratorRecipe(event, 'macerate_logs')` instead, and it'll remove the greate recipes too!

Pyritie:
I think you uploaded your configs folder by accident

Forsuin:
Sorry about that, I didn't realize that I had included it. I fixed it

Redeix (approved):

PR #1417, opened 2025-07-22
Forge, Anvil Tweaks

Zippity:
## What is the new behavior?
- Forging small gears used to create gears without any heat data. Now they do. 
- Forged small gears can also now be melted back into their component fluids. 
- A few anvil recipes had finishing sequences that were not in line with modpack convention of `[LAST HIT, SECOND LAST HIT, THIRD LAST HIT]`. These recipes have been adjusted accordingly.

## Implementation Details
- Adding heat data to small gears is done through TagPrefix matching, following modpack convention.
- Adding melting recipes to forged small gears was done by replacing the previous set of anvil recipes with a map of metals and metal tiers that generate both anvil and melting recipes for each gear material in the map. Only forged metals were included, as I was unsure if it would be useful for every small gear material to receive a melting recipe. Most of these other materials have melting temperatures higher than what a forge can achieve.

## Outcome
- All forged gears now have heat and melting data.
- All forging recipes now follow the same finishing sequence convention.

None.


Redeix (dismissed):

Pyritie:
waiting on this to move the gears into the materials.js like the others

Pyritie (approved):

Redeix (approved):

Redeix (approved):

Issue #1421, opened 2025-07-22
[🐛]: Broken Styrene Butadiene Rubber Belt Assembler Recipe


NINAustinFett:
### Was this bug reported before?

- [x] No, it was not.

### Version modpack used

0.10.1

### Environment

Forge Server

### New Worlds

Yes

### Please provide a link to your latest.log, if relevant.

N/A

### Expected Behavior

All crafting recipes yield an output of some kind

### Actual Behavior

When Styrene Butadiene Rubber Sheets are placed in an assembler set to circuit 11 the recipe yields nothing.  It appears that is supposed to yield a Styrene Butadiene Rubber belt looking at the uses for circuit 11 in the assembler. 

<img width="934" height="768" alt="Image" src="https://github.com/user-attachments/assets/709f7fe8-ffb0-4a27-9839-2697b75c424b" /> 

<img width="895" height="752" alt="Image" src="https://github.com/user-attachments/assets/ada7e409-4a9d-48ba-94a8-35e5fefa4485" /> Mousing over the output shows nothing.

### Steps to Reproduce

Search for uses of Styrene Butadiene Rubber Sheets

### Additional Information

Surely this must be a bug right?

BlueBoat29:
I found a spelling error in the recipe. It'll be fixed in my next pr

Issue #1420, opened 2025-07-22
[🐛]: Error in Rustic windmill Blades recipe


Waza46:
### Was this bug reported before?

- [x] No, it was not.

### Version modpack used

0.10.2

### Environment

Singleplayer

### New Worlds

Unsure

### Please provide a link to your latest.log, if relevant.

Not Relevant (i assume?)

### Expected Behavior

No visible recipe/Correct usable recipe

### Actual Behavior

Broken recipe shows on JEI, with a barrier block icon indicating a missing tagg

### Steps to Reproduce

press the JEI "Uses" button on burlap cloth

Search in the Crafting tab for the rustic windmill blade recipe

PR #1422, opened 2025-07-23
Feature/tier2 rocket bugfixes

BlueBoat29:
- Adds setup for sandworm mod. Devs can add the mod to their instance to test it, but I don't think we should add it to pakku until we need to
- Adds components for tier 2 rocket
    - cone, fin, rocket alloy, rocket texture
    - changed some tags to make it easier to work with
    - added liquid desh for rocket engine
    - Recipes are commented out until EV
- Cryo fluix pearl as replacement for quantum eye at EV
- Adjusted the colors on the fluix GTmaterial to better match AE2 fluix
- Bugfix for #1420 and some other empty tag recipes
- Bugfix for #1421 

idk why it says I changed the tfc lang


kubejs/startup_scripts/tfg/items.js
```diff
@@ -12,6 +12,8 @@ const registerTFGItems = (event) => {
 	event.create('tfg:unfinished_basic_electronic_circuit', 'create:sequenced_assembly')
 	event.create('tfg:unfinished_inscriber_accumulation_press')
 	event.create('tfg:vitrified_pearl')
+	event.create('tfg:cryo_fluix_pearl')
+		.translationKey('tfg:cryo_fluix_pearl')
```
Pyritie, line 16: just so you know, if the translation key is identical to the item, you don't need to specify it here

Pyritie (approved):

PR #1451, opened 2025-07-27
Universal compost items

Vazde:
## What is the new behavior?
This adds two new items: universal compost browns and greens and recipes to use them. They act as compostable items. The purpose is to give the players tools to unify distint types of compostable items to singlular types. This increases storage efficiency and allows using those distinct items in machine recipes that expect a certain amount of material.

For example the player may have 1 of each 16 fallen leaves. Alone they can't be used in a mixer, but if first crafted to universal compost they fulfill the composting recipe for that compost color.

## Outcome
Two new items: universal compost greens and browns. Recipes to convert other compostable items to them via either mortar crafting, or via forge hammer.

## Additional Information
Please note that the textures for the new items are placeholders, and should be replaced with better ones by _someone_.

We already did some review rounds of this at Discord; I hope that everything is to your satisfaction now.

It is possible that there are some recipes that conflict with either the mortar or machine recipe. Please double check.

/Vazde


kubejs/server_scripts/tfg/recipes.miscellaneous.js
Pyritie, line 1: this is missing `#tfc:compost_browns` and `#tfc:compost_greens`

kubejs/startup_scripts/tfg/items.js
Pyritie, line 1: you don't need to add `.translationKey` and `.texture` if they're just going to be in the default locations like they are here (I mean you can if you want but they effectively don't do anything)

Pyritie (changes requested):

kubejs/server_scripts/tfg/recipes.miscellaneous.js
Vazde: Added

Pyritie (approved):

Redeix (approved):

Vazde:
Realized that items which are both greens and browns can be used in the recipes. Should something be done to that, or just let it be?

Pyritie:
hmm.. I think there's only one item that's both green and brown? can't
remember
I'm not sure how you'd change it without adding a tool to the recipe (and
making it shaped) or making it dependent on crafting table position (would
result in the same questions we get all the time about small/tiny dust
piles)

PR #1468, opened 2025-07-28
More things

Redeix:
Some more gt7 model fixes and removed gtceu metal armor


Redeix:
Change for the solar panel structure

gustovafing (approved):

PR #1471, opened 2025-07-28
More Better End plants

Pyritie:
a ton of stuff copied and pasted from better end

haven't done all the block builders yet though, and only like half of these plants are for mars, I'm thinking a lot of the vines we could turn into water plants for europa?


Redeix (approved):

Redeix:
Did the emissive stuff work?

Pyritie:
uhhhhhh

PR #1487, opened 2025-07-30
Added concrete into facade materials

TheAshenWolf:
## What is the new behavior?
Between all the items that can be used to make a GregTech facade, concrete is missing.
Since wood and gravel are available, I felt like concrete, a much more stable material, should be available as well.

## Implementation Details
Just added a tag representing concrete in to tags.facades.js

## Outcome
Concrete facades \o/
image

There is no `#minecraft:concrete`, so if any future mod adds blocks tagged `#forge:concretes`, it would be added to facade list as well, which may not be desirable.

**Discord nick:**
`theashenwolf`


Pyritie (approved):

Redeix (approved):
Das conk crete

PR #1489, opened 2025-07-31
Final render fixes

Redeix:
Should be the last render fixes  and textures needed for gt7. 
The railgun receiver will need a new structure.
I think I may know a way to get the custom monitor model working with dynamic rendering. Ill try it later.


gustovafing (approved):

Pyritie (approved):
the railgun ammo texture makes me think about putting charged certus in its crafting recipe, lol

PR #1493, opened 2025-07-31
Universal compost fixes

Vazde:
## What is the new behavior?
Made universal compost crafting recipes shaped to avoid conflicts better. Also made both the crafting and machine recipes constraint out items which have both compost colors.

NOTE: The machine recipes don't currently work as they use subtract.

/Vazde


Pyritie:
implemented in #1507 instead

PR #1502, opened 2025-08-01
TFC Invar

Pyritie:
recipes.materials.js is because this is our first metal that has TFC-craftable tools, but isn't a native TFC metal, so it doesn't have things like javelins etc. (no redeix please, you don't need to make them 😩)


Redeix (approved):
Ah I see, so you added a Boolean to the materials so it should be easier to do this in the future

Pyritie:
Yeah, just add any other materials to the boolean and it'll work the same way

PR #1633, opened 2025-08-15
Feature/wan creatures and stuff

BlueBoat29:
quite a bit but mainly drops for martian creatures + spawning


defaultconfigs/wan_ancient_beasts-server.toml
Pyritie, line 1: should we buff some of the HPs a bit? since the EV level guns are available at this point

BlueBoat29: one thing is note it that most of these do have armor, so guns with more armor pen will be better

Pyritie (approved):

Redeix (approved):

Redeix:
Maybe we should make the meat look a little more alien. They already look like puppets so get weird with it

PR #1677, opened 2025-08-23
fix: Fixed Oven quest

Wicpar:
## What is the new behavior?
The oven quest still needed recently disabled Peel item, now it replaces it with the intended insulating item.

## Implementation Details
Basic config and secription change

## Outcome
Quest is now doable

## Additional Information
Still needs Russian localization, but is usable enough as-is

None


Pyritie:
Could you copy the lang string change over to [here](https://github.com/TerraFirmaGreg-Team/Tools-Modern/tree/dev/LanguageMerger/LanguageFiles/tfg/en_us/Quests) as well please? It's where all the lang strings are actually stored

Wicpar:
> Could you copy the lang string change over to [here](https://github.com/TerraFirmaGreg-Team/Tools-Modern/tree/dev/LanguageMerger/LanguageFiles/tfg/en_us/Quests) as well please? It's where all the lang strings are actually stored

Oh, just saw that you fixed it before i could

PR #1686, opened 2025-08-24
Update README.md for shaders

Pyritie:
In order to include a version of complimentary that works with TFG, we have to add stuff to our modpack description as per their license https://github.com/ComplementaryDevelopment/ComplementaryReimagined/blob/main/License.txt

### DIFF
--- a/.github/README.md
+++ b/.github/README.md
@@ -67,6 +67,8 @@ Want to help translate? Please read [this](../kubejs/README_IF_TRANSLATING.md) i
 - WithersChat from the [Monifactory](https://www.curseforge.com/minecraft/modpacks/monifactory) team for the universal circuit textures.
 - [GregTech Community Pack Modern](https://www.curseforge.com/minecraft/modpacks/gregtech-community-pack-modern) for most of their quest text.
 - [Better End](https://www.curseforge.com/minecraft/mc-mods/betterend) for some assets.
+- [Complimentary Reimagined](https://www.complementary.dev/shaders/) for their wonderful shaders. As per their [license](https://github.com/ComplementaryDevelopment/ComplementaryReimagined/blob/main/License.txt), TerraFirmaGreg-Modern accepts all responsibility for the inclusion of this shader pack in our modpack.
+- [TaidumCraft](https://modrinth.com/modpack/taidumcraft) for their modification of Complimentary Reimagined to work with Ad Astra's added dimensions.
 - Uni for Mars' dunes terrain generation.
 
 <!-- Links: -->


Xikaro:
Will we publish our version of the shaderpack on platforms?

Pyritie:
Yes, we're currently just finishing up making one. Would it be included in Modpack-Modern? or in the Shaders-Modern project, which is then added to Modpack-Modern?

Pyritie:
(I'm fine either way)

Xikaro:
I think we'll do something like a core mod, automatic publishing on cf + mr, and add it using pakku.

Xikaro:
I just think it should be done not as a setup, but as a full-fledged package.

Pyritie:
Ok in that case, the package will need something different for the complimentary license: 

>   1.3. This Pack is allowed to be redistributed as a Modified Pack as long as
>        the following conditions are met:
>          a. The Original Pack must be visibly credited in the description
>             page(s) of the Modified Pack, with a link to the official
>             Complementary Development website.
>          b. This license file must still be left inside the Modified Pack
>             without any changes being made to it, and it must be updated to
>             the latest one found in the latest version of the Original Pack
>             every time the Modified Pack gets updated to make more use of the
>             code from the Original Pack.
>          c. The Modified Pack must not have the word "Complementary" in its
>             name, and there should be at least 3 character difference between
>             the name of the Modified Pack and the Original Pack. Adding or
>             removing non-alphabetic characters or the removal of
>             "Complementary" does not count as a character difference.
>          d. The Modified Pack must look noticeably different from the
>             Original Pack in multiple common gameplay scenarios that must
>             include daytime overworld visuals, regardless of the setting or
>             variable changes in the Modified Pack and/or the Original Pack.

I'm not sure if we would qualify for (d) though, as our overworld looks basically identical. We just added compatibility for tfc fluids and blocks.
The shader is much more noticeable on other planets, though.

Pyritie:
It sounds like the license is much more permissive if we include it in the modpack directly.

Xikaro:
The problem with this approach is that the modpack will contain an archive, which is a binary file, and it will be difficult to track the changes.

Therefore, it would be better to use a fork and make changes there and publish it, I see that there are many such forks on coursforge.

Pyritie:
a fork of complimentary?

Xikaro:
Yes

Xikaro:
Or did I misunderstand your idea?

Pyritie:
alright, let's do that then

Pyritie:
alright, let's do that then

PR #1714, opened 2025-08-27
Added structures intended for the overworld world generation.

Nordic001:
## What is the new behavior?
Added additional structures based on the technology and building style of real world Norse people, that can be used in the generation of the overworld.

None


Pyritie:
@Redeix These will still need paths, loot, and jigsaws, but it's a good start, and if we can get those all done then @Nordic001 could learn from them and start doing the others by himself

Redeix:
> @Redeix These will still need paths, loot, and jigsaws, but it's a good start, and if we can get those all done then @Nordic001 could learn from them and start doing the others by himself

Okay, sounds good. I'll try to get to them soon

PR #1722, opened 2025-08-28
make popped chorus fruit edible

programmer3481:
## What is the new behavior?
makes popped chorus fruits edible


Pyritie:
put your discord name in the description if you want the contributor role btw

Issue #1740, opened 2025-09-01
[🪶]: Sheet Metal deco blocks and plate welding recipes conflict in Mechanical Press/Basin


AirRice:
### Was this feature suggested before?

- [x] No, it was not.

### Version modpack used

0.10.10

### Related problem

In the mechanical press/basin, you cannot make more than one double plate using the welding recipe as it collides with the sheet metal recipe (4 plates -> 4 of the decoration block). These blocks are unable to be recycled and thus end up wasting the material entirely

<img width="592" height="472" alt="Image" src="https://github.com/user-attachments/assets/7c52e5f7-69eb-4afa-b9dd-9673098e8388" />

<img width="585" height="412" alt="Image" src="https://github.com/user-attachments/assets/7b77cdb6-3921-40e0-acc8-0b6e1a3730ff" />

### Your solution

Give plate welding recipes a circuit number?

### Additional information

This is not a large issue as the double plate welding recipe becomes quickly obsoleted by the bender a bit down the line. Still, a bit annoying.

PR #1746, opened 2025-09-04
Fix sheet metal and welding conflict

SpicyNoodle5:
## What is the new behavior?
https://github.com/TerraFirmaGreg-Team/Modpack-Modern/issues/1740

## Implementation Details
Removed the original recipe for sheet metal blocks and they are now made in the stonecutter instead: 1 plate -> 1 sheet metal.

## Outcome
The sheet metal recipes no longer conflict with making double plates through the mechanical press.

# Additional Information
This seemed like the easiest solution to this issue, especially since most other create deco blocks can be made in a stonecutter. Open to other suggestions tho

Discord: Spicy Noodles


Pyritie:
ah I was originally gonna suggest changing it to use foil instead of plates (since it's a deco block and it's ok if those are cheap) but this can work too! unless you want it to be foil in a stonecutter?

SpicyNoodle5:
I could just 4x the output here? That'd give the same amount after converting plates to foils but with a little bit less processing

Pyritie:
sure

Pyritie (dismissed):

PR #1778, opened 2025-09-09
Item sizes

### LINKED ISSUE #767: [❓]: Why are Clay balls Medium and Heavy instead of the default of Very Small and Very Light? Is this intentional?
### Was this question asked before?

- [x] No, it was not.

### Your Question

In TFG the size and weight of clay balls are significantly increased from base TFC, to the point where the stack size is one quarter of the original. Is this an intentional change? Why was this change made?

KonSola5:
## What is the new behavior?
This PR adds item size definitions for many non-TFC items. 

## Implementation Details
- Added item size definitions to `kubejs/server_scripts/tfg/data.js` (couldn't bother with fragmenting the JS code across gazillion files).
- Added some helper tags to the `tfg` namespace to aid with item size defintions.
- Added item tags for Domum Ornamentum floating carpets and added Create Deco glass panes to `forge:glass_panes` tag.

## Outcome
- All `forge:ingots` will be Medium and Large (stacks to 16), except:
  - Clay Balls (fixes #767),
  - Bricks,
  - Nether Bricks.
- All `forge:plates` will be Medium and Large (stacks to 16), except:
  - Paper
  - Medium Density Fiberboard
  - Treated Medium Density Fiberboard
- All `forge:double_plates` will be Heavy and Large (stacks to 4).
- Wet Mud Bricks will stack to 64.
- Many blocks and items have their item sizes changed usually with these criteria:
  - Either they are made out of things that already have sizes (usually the case in case of GT items),
  - or are visually or structurally bigger than 1 block in size (usually the case in case of Create blocks),
  - or are smaller than 1/2 block in size (Framed Blocks).
- Various pipes and cables also had their sizes adjusted according to their physical size.
## Additional Information
The stack sizes should have a good balance between realism and playability, since for example a Large Water Wheel being Very Heavy and Huge would not be very fun to play. So keep that in mind and review each change thoroughly.

Also: 
- Added a Linux linked library for EffekseerNativeForJava to `.gitignore` (there already was a Windows `.dll` but there wasn't a Linux `.so`)
- For JS files that I changed, formatted them with Prettier and added type annotations to functions that take an event.

Discord nickname: KonSola5


TomPlop (commented):
Sounds good to me though I wonder if we should use this opportunity to change the weight of wires so it can be stack to 64 and maybe x2 at 64 too then x4 and x8 would be 32, finally x16 would be 16?

KonSola5:
Good idea! Pipes could get the same treatment.

Pyritie:
oh, github doesn't send notifications for when you mark as ready for review -- but this looks much better! Would you be able to move the last few tag files into kubejs as well please? 🙏

Pyritie:
also, your IDE turned all the tabs into 2 spaces, could you revert that pls

KonSola5:
What's the problem with spaces?
In the original script, tabs were 4 spaces, and they were indeed using spaces, not tab characters.

Pyritie:
the problem is that in your PR, each indentation level is now only 2 spaces instead of 4

KonSola5:
Aha, so 4 spaces are preferred here. Got it.

KonSola5:
Added a `.prettierrc` file with config for 4-wide tabs and 120 character print width so contributors that use Prettier will use these settings for formatting.

In addition: Noticed that the repository does not have consistent tab width - some files have 2-wide spaces, some 4-wide, some even mixed.

Though with this new config, every file I touched should have 4-wide tabs now.

Pyritie:
thanks! yeah some of the other 2-wide files are probably copied and pasted from elsewhere or generated -- btw did you mean to leave the other tag json files in?

KonSola5:
Good catch - forgot to remove the Domum Ornamentum floating carpet tag JSON.

Item sizes that override defaults will remain as JSON - to make sure they are 100% overriden.

KonSola5:
> Double quotes are nasty, please dont continue refactoring with them

Why so?
Is it style, or you more often use double quotes inside of strings so you don't need to escape them?

That's also configurable via Prettier, and you can set that up in the `.prettierrc` file.
The property:
`"singleQuote": true`

PR #1784, opened 2025-09-09
Bulk Salt Boiling

TreyR9:
Added Vat recipe for boiling salt water into salt, if any of the salt water-related recipes break in the future then this is the likely culprit

I was fully expecting the other salt water recipes to break but surprisingly it seems like they won't run if an invalid ingredient is present, which prevents recipe conflicts.


TreyR9:
oops lmao, I copied it straight from my personal instance, feel free to nuke that

PR #1790, opened 2025-09-10
Small tweaks for progression

TomPlop:
-Adjust the Field Generators to follow your new path of progression on Pearl
-Adjust some backpacks upgrades so they are on the same level as there circuits
-Update the crafting recipe of the quantum eye for cryo pearl
-Bump of one tier Desh and Ostrum recipes
-Make the Power Substation (PSS) available at EV


Pyritie (approved):
can the substation casing be recycled? we should remove or fix that if so, otherwise you can get very easy iridium

TomPlop:
> can the substation casing be recycled? we should remove or fix that if so, otherwise you can get very easy iridium

Good catch you are totally right it can be lol

PR #1833, opened 2025-09-15
Jute net to be made with Flax

### LINKED ISSUE #1815: [🪶]:

- [x] No, it was not.

### Version modpack used

0.10.12

### Related problem

Flax in general appears to be interchangeable with jute for most recipes, but you need specifically jute to make the jute net for a fishing net, which seems like an oversight

### Your solution

Allow the fishing net to be crafted with a new flax net item, crafted in the same way as a jute net, but with flax

### Additional information

_No response_

Nashy1232:
## What is the new behavior?
Add Flax as a valid option for the Jute Net

## Implementation Details
Replace input `tfc:jute_net` for `#tfg:burlap_fiber`

## Outcome
fix #1815

### DIFF
--- a/kubejs/server_scripts/tfc/recipes.js
+++ b/kubejs/server_scripts/tfc/recipes.js
@@ -334,4 +334,9 @@ const registerTFCRecipes = (event) => {
     // jute net -> burlap net
     event.replaceInput({ id: 'tfc:crafting/jute_net'}, 'tfc:jute_fiber', '#tfg:burlap_fiber')
 
+    // horse armor to use burlap
+    global.TFC_EQUIPMENT_METALS.forEach(material => {
+        event.replaceInput({ id: `tfc:crafting/${material}_horse_armor`}, `tfc:jute_fiber`, `#tfg:burlap_fiber`)
+    })
+
 }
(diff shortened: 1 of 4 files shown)


Nashy1232:
whooopsie, I staged the wrong file, this has been corrected.

Pyritie:
yeah, recipe ID is better where possible (or just make a second recipe altogether)

btw I think I just remembered that there's another recipe that uses jute by itself - the firmaciv rope coil. Would you be able to add a recipe for that too?

Nashy1232:
I extended the scope of this merge, so i'll add some details

## What is the new behavior?
Add Flax as a valid option for the Jute Net
Add Flax as a valid option for metal horse armor



## Implementation Details
Replace input `tfc:jute_fiber` for `#tfg:burlap_fiber`

Added new global containing the metals used to make equipment
Updated existing `kubejs/server_scripts/tfc_grooming_station/recipes.js` to use this global

## Outcome
fix #1815 
fix allowing horse armor to be created using either jute or flax

Pyritie (approved):

TomPlop (approved):

PR #1854, opened 2025-09-17
Add catwalk facades

programmer3481:
## What is the new behavior?
Adds catwalk facades (both tfg and ae2)
also fix being able to craft singular bricks into gregtech stone facades

## Implementation Details
tfg's facade tag is an item tag while ae2's facade tag is a block tag so i just put the ae2 facade tag adding logic into the ae2 folder and basically copied tfg folder's as a block tag
also just specified each of the mud brick blocks manually since there was no common tag just for them

## Outcome
catwalk facades added
also no longer able to craft gregtech stone facades out of singular tfc mud bricks

## Additional Information

discord: darkmatter3481

### DIFF
--- a/kubejs/server_scripts/ae2/tags.facades.js
+++ b/kubejs/server_scripts/ae2/tags.facades.js
@@ -2,19 +2,9 @@
 "use strict";
 
 function registerAe2FacadeWhitelistTags(event) {
-    
-    const createdeco_catwalks_fm =
-	[
-		'andesite',
-		'brass',
-		'iron',
-		'copper',
-		'industrial_iron',
-		'zinc'
-	]
 
     //createdeco catwalks
-	createdeco_catwalks_fm.forEach(facade_material => {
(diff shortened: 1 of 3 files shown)


kubejs/server_scripts/ae2/tags.facades.js
```diff
@@ -0,0 +1,20 @@
+// priority: 0
+"use strict";
+
+function registerAe2FacadeWhitelistTags(event) {
+    
+    const createdeco_catwalks_fm =
```
Pyritie: might be worth adding this to startup_scripts/create_deco/constants.js instead of duplicating it twice?

programmer3481: done

BlueBoat29 (approved):

BlueBoat29:
add discord name if you want contributor role :)

Pyritie (approved):

PR #1887, opened 2025-09-24
add evaporation tower and nuclear turbine

TomPlop:
Adding textures and disabling Evaporation Tower from Kubejs


Redeix (approved):

TomPlop:
I use the Biovat texture right now but that could be change I just wanted it to work at least

PR #1892, opened 2025-09-25
Create packager recipe buffs

AirRice:
## What is the new behavior?
Packagers and Bottlers have their costs reduced by making them require metal casings rather than ULV Hulls.

## Implementation Details
Playing TFG with my friends we found out that the ULV hulls that are part of the packager/bottler recipes are prohibitively expensive, especially early game. With them being so crucial to the setting up of a logistics network, this PR addresses that by making the recipe require a casing rather than a hull (remove rubber dependency) and giving 2x returns for the crafts.

Additionally, the LV assembler cardboard recipe has also been buffed a bit to require less paper and glue because it's a tedious recipe to begin with. This should make it a lot less annoying to make.

Stock links/factory gauge recipes are left unchanged, as the automation of parts needed in them is a core mechanic of LV/MV and they will naturally become less tedious to craft as an LV/MV factory expands

Original PR proposed replacing the most expensive part of these recipes with metal (andesite) casings, cutting down the wrought iron usage to nearly 10% of the original cost, before it was revised to current state.

## Outcome
Changes the recipes for the create Packager, create factory logistics Bottler, and cardboard in the LV assembler. 
Overall this should make these feel better to craft.

N/A

### DIFF
--- a/kubejs/server_scripts/create/recipes.js
+++ b/kubejs/server_scripts/create/recipes.js
@@ -1732,12 +1732,6 @@ const registerCreateRecipes = (event) => {
 		F: '#tfg:metal_bars'
 	}).id('tfg:create/shaped/packager')
 
-	event.recipes.gtceu.assembler('tfg:create/packager')
-		.itemInputs('gtceu:ulv_machine_casing', '4x #forge:rods/wrought_iron', '#forge:springs/wrought_iron', '4x create:cardboard', '#gtceu:circuits/ulv')
-		.itemOutputs('2x create:packager')
-		.duration(200)
-		.EUt(20)
-
 	event.shaped('create:item_hatch', [
 		'A',
 		'B',
--- a/kubejs/server_scripts/create/recipes.js
+++ b/kubejs/server_scripts/create/recipes.js
@@ -1712,8 +1712,8 @@ const registerCreateRecipes = (event) => {
 	}).id('tfg:create/shaped/cardboard_from_glue')
 
 	event.recipes.gtceu.assembler('tfg:create/cardboard_from_glue')
-		.itemInputs('5x minecraft:paper')
-		.inputFluids(Fluid.of('gtceu:glue', 100))
+		.itemInputs('4x minecraft:paper')
+		.inputFluids(Fluid.of('gtceu:glue', 50))
 		.circuit(5)
 		.itemOutputs('2x create:cardboard')
 		.duration(200)
@@ -1726,14 +1726,14 @@ const registerCreateRecipes = (event) => {
 	], {
 		A: '#forge:rods/wrought_iron',
 		B: '#forge:springs/wrought_iron',
-		C: 'gtceu:ulv_machine_hull',
+		C: 'create:andesite_casing',
 		D: 'create:bound_cardboard_block',
 		E: 'create:electron_tube',
 		F: '#tfg:metal_bars'
 	}).id('tfg:create/shaped/packager')
 
 	event.recipes.gtceu.assembler('tfg:create/packager')
-		.itemInputs('gtceu:ulv_machine_hull', '4x #forge:rods/wrought_iron', '#forge:springs/wrought_iron', '4x create:cardboard', '#forge:string', '#gtceu:circuits/ulv')
+		.itemInputs('create:andesite_casing', '4x #forge:rods/wrought_iron', '#forge:springs/wrought_iron', '4x create:cardboard', '#gtceu:circuits/ulv')
 		.itemOutputs('create:packager')
 		.duration(200)
 		.EUt(20)
--- a/kubejs/server_scripts/create/recipes.js
+++ b/kubejs/server_scripts/create/recipes.js
@@ -1719,22 +1719,22 @@ const registerCreateRecipes = (event) => {
 		.duration(200)
 		.EUt(7)
 
-	event.shaped('create:packager', [
+	event.shaped('2x create:packager', [
 		'AAA',
 		'BCD',
 		'EFE'
 	], {
 		A: '#forge:rods/wrought_iron',
 		B: '#forge:springs/wrought_iron',
-		C: 'create:andesite_casing',
+		C: 'gtceu:ulv_machine_casing',
 		D: 'create:bound_cardboard_block',
 		E: 'create:electron_tube',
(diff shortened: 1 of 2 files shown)


kubejs/server_scripts/create_factory_logistics/recipes.js
```diff
@@ -28,17 +28,18 @@ function registerCreateFactoryLogisticsRecipes(event) {
 	], {
 		A: '#forge:rods/copper',
 		B: '#forge:springs/copper',
-		C: 'gtceu:ulv_machine_hull',
+		C: 'create:andesite_casing',
```
Pyritie: what about copper casing? since that's the traditional create fluid-related casing

kubejs/server_scripts/create/recipes.js
```diff
@@ -1726,14 +1726,14 @@ const registerCreateRecipes = (event) => {
...
-		C: 'gtceu:ulv_machine_hull',
+		C: 'create:andesite_casing',
 		D: 'create:bound_cardboard_block',
 		E: 'create:electron_tube',
 		F: '#tfg:metal_bars'
 	}).id('tfg:create/shaped/packager')
 
 	event.recipes.gtceu.assembler('tfg:create/packager')
-		.itemInputs('gtceu:ulv_machine_hull', '4x #forge:rods/wrought_iron', '#forge:springs/wrought_iron', '4x create:cardboard', '#forge:string', '#gtceu:circuits/ulv')
+		.itemInputs('create:andesite_casing', '4x #forge:rods/wrought_iron', '#forge:springs/wrought_iron', '4x create:cardboard', '#gtceu:circuits/ulv')
```
TomPlop: Why removing the string also? It's a good way to incentivize farming at that stage.

kubejs/server_scripts/create/recipes.js
```diff
@@ -1726,14 +1726,14 @@ const registerCreateRecipes = (event) => {
 	], {
 		A: '#forge:rods/wrought_iron',
 		B: '#forge:springs/wrought_iron',
-		C: 'gtceu:ulv_machine_hull',
+		C: 'create:andesite_casing',
```
TomPlop: Going from ulv machine hull to andesite casing is wild. We could nerf it to ULV Machine Casing and not Hull so players who don't know how to make plastic at LV aren't too penalized.

ULV Machine Casing is the one use by most of Create so it does make sense and it's extremely cheap at that point of the game. Either you have a Steam bloomery or an Arc furnace, both cases are easy and really should be the minimum automation required to begin using an automation mod.

kubejs/server_scripts/create_factory_logistics/recipes.js
TomPlop: Same thought using the machine casing and not the hull. Using Copper or Andesite casing is wild.

kubejs/server_scripts/create/recipes.js
Pyritie: Mm I dunno, you do want to encourage people to actually use automation instead of thinking "oh that looks too expensive" and then skipping it entirely, making their experience more miserable

TomPlop: Then you get players that put 3 pattern providers on the same wiremill and they quit the game because the leap from working to feel the reward from everything is free is too much and their base isn't ready.

I agree that setting up plastic at LV even though it's damn easy can set a toll on new players, but setting up a steam bloomery with a helve hammer? Come on.

I would prefer then that you require a ULV Machine Hull but get two Packagers or Bottler.

Still a 40% nerf on cost. But at least please passive your wrought iron or you gonna quit the game.

Pyritie: I would be more ok with that sort of thing. I agree that the hull is probably too much at the moment.
I had the same thought with the gauges being kinda expensive to make too. Maybe they could also return two each.

TomPlop: Gauge are definitely expensive but they already give two. The thing with Gauge is that they are a crazy strong item, it's basically a ME Requester from AE2. I advise strongly against making them easier to get as their promote a dangerous way to play. Also it's the endgame item of Create Logistics.

Less machine and less passive.

I think it's the good compromise to ask for ULV Machine Casing and give two Packager/Bottler.

kubejs/server_scripts/create/recipes.js
AirRice: The string was there because the bound cardboard block requires one, but going by the idea of assembler recipes usually doing away with 'binding' components like screws in other recipes, I think this makes sense. The original recipe does still require string, this is just a small cut to costs starting in LV

kubejs/server_scripts/create/recipes.js
AirRice: Mm, looking back given that the create machines crafted around that period (vac chamber, deployer, etc) require ULV casings as well I think that's a fair baseline to balance against. The issue with hulls was that they require rubber plated cables to make, which unnecessarily delayed getting them
Given that and the large amounts of paper needed which wasn't really fully addressed in the original PR I think I'll make the packager and bottler take a ulv casing rather than a metal casing but return 2x per craft.

I think the gauge recipes are fine, given that they already return 2x, and also what Tom said above

kubejs/server_scripts/create/recipes.js
TomPlop: There is no strings in the crafting recipe though? Actually they are pretty different. Also why do we have an Assembler recipe and a Crafting recipe even though there is no tools in the crafting recipe.

Some weird stuff going on.

kubejs/server_scripts/create/recipes.js
TomPlop: sounds good

kubejs/server_scripts/create/recipes.js
TomPlop: Oh the string is in the cardboard

Pyritie (approved):

Pyritie:
oh I noticed that the shaped recipe wants electron tubes but the assembler one wants ulv circuits, could you make those consistent too pls

TomPlop:
Yeah that’s what I was saying! Just remove the Assembler recipe. We have to stop making assembler recipe if there is no tools used.

TomPlop (approved):

PR #1903, opened 2025-09-26
flux dupe fix

Redeix:


Pyritie:
You could make the recipe check for the nbt of the item instead of removing it

Redeix:
Yeah, but that's like a bunch of work 😶‍🌫️

PR #1962, opened 2025-10-06
Fixed a Linux crash caused by a mispressed space in Greate removes

Xaligal:
## What is the new behavior?
Linux no longer freaks out trying to run the pack

## Implementation Details
Shamefully pressed backspace in the greate remove

## Outcome
Linux no longer freaks out trying to run the pack


BlueBoat29 (approved):

SunderSoul:
oop didn't think me messing with the metal bars would mess with this pr, sorry

PR #1970, opened 2025-10-06
all pressure plates to circ 3

Nashy1232:
## What is the new behavior?
changed all pressure plate assembler recipes to use circuit 3 to avoid most conflicts, issue stemmed from pressure plates needing no circuit and also accepting any spring. Some recipes such as creates precision mechanism requires a gold plate and a gold spring, this was conflicting with "light weighted pressure plate"


Nashy1232:
I'd appreciate if someone double checked for me to make sure circ 3 isnt going to be an issue

Redeix (approved):
Code looks solid. Can't test in game since I'm at work, but if someone else wants to review after testing for conflicts.

PR #1978, opened 2025-10-07
First fixes for Mars Fission

TomPlop:
-Switch Nuclear and Uranium mistake
-Increase Tritiated Water so it makes exactly 9 Nano Wafer

-Fix the different multiblocks air/any

### DIFF
--- a/kubejs/startup_scripts/gtceu/machines.js
+++ b/kubejs/startup_scripts/gtceu/machines.js
@@ -482,7 +482,7 @@ const registerGTCEuMachines = (event) => {
 			.aisle('KKAAAAAAA', 'AAAAAAAAA', 'AAAAAAAAA', '         ', '         ' )
 			.aisle('BAAAAAAAA', 'B       D', 'BBBBBBBAA', ' BCCCB   ', ' BBBBB   ' )
 			.aisle('AAAAAAAAA', 'A       D', 'BB   BBAA', ' C   C   ', ' BIAHB   ' )
-			.aisle('BEBEBEAAA', 'BFBFBFA#D', 'BBBBBBBAA', ' BCCCB   ', ' BBBBB   ' )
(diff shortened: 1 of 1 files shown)


kubejs/startup_scripts/gtceu/machines.js
```diff
@@ -481,14 +482,15 @@ const registerGTCEuMachines = (event) => {
 			.aisle('KKAAAAAAA', 'AAAAAAAAA', 'AAAAAAAAA', '         ', '         ' )
 			.aisle('BAAAAAAAA', 'B       D', 'BBBBBBBAA', ' BCCCB   ', ' BBBBB   ' )
 			.aisle('AAAAAAAAA', 'A       D', 'BB   BBAA', ' C   C   ', ' BIAHB   ' )
-			.aisle('BEBEBEAAA', 'BEBEBEA#D', 'BBBBBBBAA', ' BCCCB   ', ' BBBBB   ' )
+			.aisle('BEBEBEAAA', 'BFBFBFA#D', 'BBBBBBBAA', ' BCCCB   ', ' BBBBB   ' )
```
Pyritie: three item inputs?

TomPlop: That's just the shapeinfo

Pyritie: so? why three item inputs, you only ever need one, and you know people like to blindly follow what the shapeinfo says

TomPlop: Well you need only one fluid input hatch then. If someone blindly follows the ShapeInfo at EV we are fuck because the machine may not run 😂

I don’t think anyone is that dumb at that point but give me a minute I will fix it.

Pyritie (approved):

PR #1986, opened 2025-10-09
Add "Recipes.Final" script

Nebby1999:
-# no way, i'm making a contribution :o
## What is the new behavior?

While i was playing, i noticed that new Scraping recipes where added (Flax), however, i could also notice that some of the new scraping recipes didnt have automatic implementations on Create using Sequenced Assembly.

From what i could tell, the original recipe utilized all the registered scraping recipes, to create equivalents using create sequenced assembly + a tfc knife. However, this recipe was running prior to TFG registering its own recipes, causing the contradiction.

As a result, i created a new file called "recipes.final", ideally any recipe that acts upon existing recipe types to create equivalents should be here, as this will be called at the end of the Recipe registration procedure

## Implementation Details

Added a new file which it's method gets called at the end of the registration method

## Outcome

All recipes that use scrapping can now be worked on a Sequenced Assembly setup.

## Additional Information

image

### DIFF
--- a/kubejs/server_scripts/create/recipes.js
+++ b/kubejs/server_scripts/create/recipes.js
@@ -1676,16 +1676,6 @@ const registerCreateRecipes = (event) => {
 			.EUt(7)
 	})
 
-	//Allow automatic scraping by using sequenced assembly
-	event.forEachRecipe({ type: 'tfc:scraping' }, r => {
-		let originalRecipeIngredient = r.json.get("ingredient").get("item");
-		let output = r.originalRecipeResult;
-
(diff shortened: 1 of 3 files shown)


Metriximor:
I was just about to open a PR for this but you beat me to it. Nice work! :)

Pyritie:
I mentioned this in chat earlier, but this isn't a good idea because create treats sequenced assembly recipes with multiple outputs as possible _different_ results. (That's how the "scrap" results work normally)
Scraping, on the other hand, can have multiple outputs.

PR #2001, opened 2025-10-10
Meats and eats

Redeix:
Added meats for cruncher, surfer, primitive creatures, and the stackatick (couldnt resist the pun).
Updated textures for the goober and springling.


Redeix:
Im really proud of myself for getting the stick'a'stackatick pun to work in portuguese too

Pyritie:
oh yummers

should we recolor the other mars meats too?

Redeix:
Yeah, I might go through the other ones sooner or later. Just wanted to get this batch in now before I get distracted

Pyritie:
gamestar made all of the tfg ones so I think just hue shifting then is fine? (and now we have a spare meat texture from the old goober one!)

PR #2007, opened 2025-10-11
add conveyor assembler recipe

Nashy1232:
## What is the new behavior?
recipe for assembler to make chain conveyor


Pyritie:
are you sure this doesn't need a circuit number? it feels very recipe collision-y

Nashy1232:
> are you sure this doesn't need a circuit number? it feels very recipe collision-y

good catch, i was supposed to put it on 3 to go with all the other create stuff

Pyritie:
no that's the opposite, it should be on a _different_ circuit so it doesn't collide with the other create recipes

Nashy1232:
> no that's the opposite, it should be on a _different_ circuit so it doesn't collide with the other create recipes

hmm as far as i can tell it doesnt, but its hard to tell at times, so, i'll use 20, it has the create frogport on 20 and create roller and some doors, only 8 assembler recipes on circ 20 and none conflict.

Pyritie:
that sounds good then!

PR #2013, opened 2025-10-11
Impure Dust Washing in Barrels

dimethylene:
I was annoyed in game that I needed to walk 50 or so blocks to wash my dusts, so this is a small addition to allow impure dusts to be washed via an instant barrel recipe.

### DIFF
--- a/kubejs/server_scripts/tfg/recipes.miscellaneous.js
+++ b/kubejs/server_scripts/tfg/recipes.miscellaneous.js
@@ -1459,13 +1459,4 @@ function registerTFGMiscellaneousRecipes(event) {
 		.duration(300)
 		.EUt(GTValues.VA[GTValues.EV])
 	
-	Ingredient.of("#forge:impure_dusts").getItemIds().forEach(itemid => {
-		event.recipes.tfc.barrel_instant()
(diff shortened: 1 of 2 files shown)


kubejs/server_scripts/tfg/recipes.miscellaneous.js
TomPlop, line 1: wasnt itemid crashing kubejs ?

dimethylene: it worked fine on my end

Pyritie: only if you left the namespace in when doing the recipe ID, tom

dimethylene: i would do purified ores as well but some of them are from create and have a different itemid

dimethylene:
if the material system means i don't have to wrangle the itemids with replace, then absolutely

Pyritie:
I'd suggest looking at gregtech/recipes.material.js, finding the functions for crushed/impure ore, and then looking at how the other code uses materials and putting your barrel recipes in there

PR #2005, opened 2025-10-10
[Molden] Stone tfgfication pt3.5

vidal-adrien:
## What is the new behavior?
Chiseling support for Create stones, Quartz and Certus Quartz. Fixes block chiseling conflicts.

## Outcome
- Adds 2 Quartz Walls.
- Adds Chiseling and Macerator recipes for Quartz shaped blocks (stairs/slab/wall) + some shaped recipes for pillars. 
- Adds Chiseling and Macerator recipes for Certus Quartz shaped blocks + some shaped recipes for pillars.
- Adds Chiseling and Macerator recipes for Create decorative stone shaped blocks + some shaped recipes for pillars. Further connects some create stone block sets to their TFC or vanilla counterparts.
- Removes all block -> wall chiseling recipes, solving the conflict with block -> block recipes
- Reverts changes to stonecutting that made some stone blocks give 2 wall blocks. This change was intended to match the 1 block -> 2 slabs -> 2 walls ratio available through hand chiseling but the amount of recipes to edit to bring this change to all blocks would be insane.

PR #2017, opened 2025-10-12
Stone tfgfication pt3.5

vidal-adrien:
## What is the new behavior?
Continuation of https://github.com/TerraFirmaGreg-Team/Modpack-Modern/pull/2005. Apparently not all intended commits made it to that PR. Additionally:
- Adds GT packing recipes for gravel from loose stone based on https://github.com/TerraFirmaGreg-Team/Modpack-Modern/pull/2015
- Adds cobble unpacking recipes for non-TFC stones.
- Fixes that blackstone loose stones glued into raw blackstone instead of cobble.

## Implementation Details
Packing from loose to cobble was not added yet. Will reconsider when undoing the loose <-> gravel thing later.


dimethylene:
i wasn't too sure about the cobble recipe either, but nobody uses the assembler recipe (with concrete) because rock breakers exist and are faster

PR #2029, opened 2025-10-13
additional pineapple fiber recipe

Nashy1232:
## What is the new behavior?
Additional recipe to make pineapple fiber, currently only one recipe exists, craft a pineapple with a knife, additional recipe that i added is using a pineapple on a GTCEU Cutter

### DIFF
--- a/kubejs/server_scripts/firmalife/recipes.js
+++ b/kubejs/server_scripts/firmalife/recipes.js
@@ -108,13 +108,6 @@ const registerFirmaLifeRecipes = (event) => {
 		.duration(50)
 		.EUt(2)
 
-    // Pineapple fiber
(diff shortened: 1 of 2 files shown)


Pyritie:
this should be in the food processor instead, because normally this requires the pineapple to be dried first

Nashy1232:
> this should be in the food processor instead, because normally this requires the pineapple to be dried first

thoughts on making it include nitrogen to mimic the drying process?

Pyritie:
that would make sense, though check that it doesn't overlap with the existing drying recipe (so, different circuit)

Nashy1232:
I dont see any conflicts with circuit 3, should be good.

javaw_F0fzmv3Upb

Pyritie (approved):

PR #2043, opened 2025-10-16
Normalize ladders as fuel

oosyrag:
## What is the new behavior?
Ladders as fuel was an outlier in value due to the volume of ladders created from lumber.
image

## Implementation Details
Doubled ladders output from sticks recipe from 4->8
Halved ladders output from lumber recipe from 16->8
One more lumber needed in recipe to match crafting pattern from sticks
Reduced burn time of ladder and wood frames in blaze burner by 1/3
Removed ladders and wood frames as valid fuel for steam boilers (couldn't figure out how to change the duration of these)

## Additional Information

```
event.recipes.gtceu.large_boiler('gtceu:large_boiler/minecraft_ladder')
		.duration(9)
event.recipes.gtceu.steam_boiler('gtceu:steam_boiler/minecraft_ladder')
		.duration(1200)
event.recipes.gtceu.large_boiler('gtceu:large_boiler/gtceu_wood_frame')
		.duration(21)
event.recipes.gtceu.steam_boiler('gtceu:steam_boiler/gtceu_wood_frame')
		.duration(2400)
```
The above is what I tried and the numbers I would have used for the boilers (reduced from current values by 1/3), but it didn't work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 ## [Unreleased]
 ### Changes
+- Normalized ladder crafting outputs, removed ladders as steam boiler fuel @oosyrag
 ### Bug fixes
 ### Translation updates
 


Pyritie (approved):

Redeix (approved):
I don't really like the solution but I guess if it works

oosyrag:
I'll keep fiddling with this and add the recipes back with lowered durations when I figure it out.

PR #2059, opened 2025-10-19
Unify emitter recipes

SpicyNoodle5:
## What is the new behavior?
The field generator recipes were inconsistent between the crafting table and assembler

## Implementation Details
Added replaceInputs for the assembler recipes to match the crafting table recipes of field generators

Discord: Spicy Noodles

### DIFF
--- a/kubejs/server_scripts/gregtech/recipes.js
+++ b/kubejs/server_scripts/gregtech/recipes.js
@@ -1576,12 +1576,42 @@ const registerGTCEURecipes = (event) => {
 		D: 'gtceu:palladium_substation'
 	}).addMaterialInfo().id('gtceu:shaped/power_substation')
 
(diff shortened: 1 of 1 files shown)


Pyritie:
can these be recycled? because if they can, I'm not sure the replaceInput recipes were a good idea in the first place

SpicyNoodle5:
They do recycle but only to the wire/plate materials used
image

SpicyNoodle5:
Actually nvm, the EV one gives nether star dust in a macerator

Pyritie:
figures -- to fix that, these will need shaped recipes with the same recipe ID as the originals (with `event.recipes.gtceu.shaped` instead of just `event.shaped`, then add `.addMaterialInfo()` at the end of the recipe (Ctrl+F for this to see how other recipes do it). That'll update the recycling recipes, and then you can leave the assembler ones alone too.

Would you be able to do this?

TomPlop:
That's my bad always forgetting about the recycle recipes.

SpicyNoodle5:
Updated, used the MV field generator as a reference

PR #2060, opened 2025-10-19
Adjust claims

Pyritie:
Adjusting some of the claim limits because more of our space progression requires having machines set up in different dimensions in different places, and a limit of 3 isn't going to work with that. Are these numbers good do you think?

Unrelated but someone said they couldn't claim chunks in the nether/beneath, do you know why that would be? I don't see it in any files


Xikaro:
You can change it, thanks for the warning, for a single game, you can set any values.
For the server, this encourages them to team up rather than play alone to increase the number of chunks

Pyritie:
Yeah I just wanted to check since I know the server is a potato lol

PR #2065, opened 2025-10-20
fix some recipe issues in EV TACZ extended mags

BlueBoat29:
I HAVE NOT TESTED THIS 

it all should still work but don't approve until someone tests it


Pyritie:
I'm gonna merge and then test

Pyritie:
works for me!

image

PR #2067, opened 2025-10-20
Mars sandstorms

Pyritie:
Climate stuff for the sandstorms. They just go on a sin wave for now, but this is good enough to start with since I am bad at math


Redeix (dismissed):

kubejs/server_scripts/tfg/events.mars.js
BlueBoat29, line 1: could add a comment saying that this is all debug code

Pyritie: good idea

PR #2076, opened 2025-10-22
Server overrides

Pyritie:
- Changed claim limits for servers a little
- Changed claim limits for single player a lot
- Changed tfc food decay modifier to 0.25 for servers
(I also changed the bucket thing in the single player tfc-server.toml too)

### DIFF
--- a/.pakku/server-overrides/defaultconfigs/ftbranks/ranks.snbt
+++ b/.pakku/server-overrides/defaultconfigs/ftbranks/ranks.snbt
@@ -5,7 +5,7 @@
 		condition: "always_active"
 		ftbranks.name_format: "&a{name}"
 		ftbchunks.max_claimed: 50
-		ftbchunks.max_force_loaded: 8
+		ftbchunks.max_force_loaded: 6
 		ftbchunks.chunk_load_offline: false
 		command.trashcan: true
 		command.enderchest: false
@@ -15,7 +15,7 @@
 		power: 50
 		ftbranks.name_format: "&e{name}"
 		ftbchunks.max_claimed: 200
-		ftbchunks.max_force_loaded: 10
+		ftbchunks.max_force_loaded: 8
 		ftbchunks.chunk_load_offline: true
 	}
 	moderator_girl: {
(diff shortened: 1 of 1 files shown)


defaultconfigs/ftbranks/ranks.snbt
Xikaro, line 1: Here you can remove the ranks altogether, leave only the member

.pakku/server-overrides/defaultconfigs/ftbranks/ranks.snbt
Xikaro, line 1: Are such values really necessary here?
I adjusted them a little earlier, but I don't see the point in a large number, since the players in the team have a number of chunks added together, this should compensate for the small number, but also force the players to join teams.

defaultconfigs/ftbranks/ranks.snbt
Pyritie: done

.pakku/server-overrides/defaultconfigs/ftbranks/ranks.snbt
Pyritie: Right now, if a player wants to do all of the infinites, they would need to force load:
- at least 1 chunk on the overworld
- at least 4 chunks on the moon
- at least 2 chunks on mars

If they don't care about the moon infinites, then 4 would be doable but awkward.
I could understand maybe changing this to 6 or 7, but 4 still feels too low to be enjoyable? We get a lot of questions about how to work around the server's strict limits, but if it does cause too much lag on the server then I can understand

Xikaro: Let's check these parameters, yes, it loads the server primarily because of the newcomers, as they come in, set up privates and downloads, and no longer appear on the server, the shutdown has already been done, after like a month
But we have to survive this month.

If the issue is the need for a large number of chunks at a later stage, I once had the idea to implement rank allocation based on time spent on the server, which would cover the number of chunks and eliminate those who are fleeting on the server.

Pyritie: that could also work!
maybe ftbquests has something where completing a quest would also give you more chunks? so you'd get more chunks as you reach new tiers.

Xikaro: Yes, I've seen that too

PR #2088, opened 2025-10-23
Lightning rod quest

theMegaTech:
Adds a lightning rod quest, seeing as there are basically daily complaints about lightning burning someones wooden stuff and how surprisingly little amount of people even knows that the lightning rod actually exists and is good :3


Pyritie:
you'll need to change your text to use {lang IDs} like the other quests, then put the english text into the Tools-Modern repo so our translators can translate it

PR #2096, opened 2025-10-24
Update recipes.miscellaneous.js

SverhRazum-Nah:
Polycap and phantom string and fabric was on circuit 0 with zero reason

image


Pyritie (approved):

Redeix (approved):
Who would have set it to that? 🙂

PR #2136, opened 2025-10-31
Updated lv_low_voltage.snbt

reset300:
Fixed Cupronickel dependencies

Cupronickel ingot in LV Chapter dont have dependencies

Cupronickel coil in LV Chapter now have dependencies:
-Cupronickel Ingot
-Basic Liquid Extractor 

Also Moved Chemical reactor, liquid solidifier and rotor mold so it will not mess up
image


TomPlop:
image
This is how it is right now. I don't know if the change is needed but why not.

Pyritie:
I mean I guess it makes sense

Pyritie (approved):

PR #2158, opened 2025-11-04
pretty much all the molecule drawings for our current stuff

BlueBoat29:
can't add the mod to pakku yet since this was made under an indev mod version, but it should be fine to sit here.

MAKE SURE YOU ACTUALLY THOROUGHLY REVIEW IT :}


Redeix:
Btw has the mod already been added to pakku-lock?

Pyritie:
> Btw has the mod already been added to pakku-lock?

@\grok did he read the OP?

PR #2186, opened 2025-11-07
Surely this is the last one

BlueBoat29:
more molecule changes


Pyritie (dismissed):

config/moldraw.yaml
```diff
@@ -0,0 +1,14 @@
+enabled: true
+
+debugMode: false
+
+coloredAtoms: true
+
+useMaterialColors: false
```
Pyritie, line 7: 🥺🙏

BlueBoat29: its just a config 💀

PR #2210, opened 2025-11-12
Zirconium stuff

Pyritie:
- process to turn zircon into zirconium ingots (similar to titanium loop, encouraging people to move it to venus along with their oreproc)
- process to create [zirconium diboride](https://en.wikipedia.org/wiki/Zirconium_diboride) using bromine, which is for the HSS-Z coil 
- part of that also involved [boron carbide](https://en.wikipedia.org/wiki/Boron_carbide), another interesting material

I haven't tested any of the chemistry yet to check that it loops properly

this can wait until after 0.11.8

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,12 +3,16 @@
 ### Changes
 - Add shapeless p2p recipes (#2234) @SpicyNoodle5
 - Added assembler, smelting and heating recipes for firmalife ovens (#2220) @SpicyNoodle5
-- Raised burning fluid threshold from 370K to 1300K @Redeix
+- Raised player-burning fluid threshold from 370K to 1300K @Redeix
 - Added partial unification to the gtceu wax material @Redeix
+- Mars endermen can no longer pick up the wrong kind of warped/crimson sapling, and the correct kind has been added to their loot tables (#2240)
 ### Bug fixes
 - Fixed brown gravy recipe @Redeix
 - Fixed instant mac recipe requiring less cardboard than it gives back @Redeix
 - Fixed Ad Astra moon cheese nutrition values being inconsistent @Redeix
+- Fixed some quest typos @Pyritie
+- Fixed duplicate bed dyeing recipe (#2229) @Pyritie
+- Fixed Basic Hermetic Casing missing a mineable tag and loot table (#2221) @Pyritie
 ### Translation updates
 
 ## [0.11.8] - 19-11-2025
--- a/kubejs/assets/tfc/patchouli_books/field_guide/en_us/entries/getting_started/primitive_alloys.json
+++ b/kubejs/assets/tfc/patchouli_books/field_guide/en_us/entries/getting_started/primitive_alloys.json
@@ -14,7 +14,7 @@
 		},
 		{
 			"type": "patchouli:text",
-			"text": "For example, to create 1000 mB of $(thing)Bronze$() (shown to the right), you would need between 880 and 920 mB of $(thing)Copper$(), and between 80 and 120 mB of $(thing)Tin$().$(br2)The next three pages show the recipes of the three bronzes. Each type of bronze can be used to make tools, $(l:mechanics/armor)Armor$(), and other metal items. They are slightly different so resulting tools will have different durability, efficiency, and attack damage."
+			"text": "For example, to create 1000 mB of $(thing)Bronze$() (shown to the right), you would need between 700 and 800 mB of $(thing)Copper$(), and between 200 and 300 mB of $(thing)Tin$().$(br2)The next three pages show the recipes of the three bronzes. Each type of bronze can be used to make tools, $(l:mechanics/armor)Armor$(), and other metal items. They are slightly different so resulting tools will have different durability, efficiency, and attack damage."
 		},
 		{
 			"type": "patchouli:spotlight",
--- a/kubejs/assets/tfc/patchouli_books/field_guide/zh_cn/categories/beneath.json
+++ b/kubejs/assets/tfc/patchouli_books/field_guide/zh_cn/categories/beneath.json
@@ -1,0 +1,6 @@
+{
+    "name": "下界……？",
+    "description": "关于幽冥之地的一切",
+    "icon": "minecraft:netherrack",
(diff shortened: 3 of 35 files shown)


Pyritie:
ready for review

TomPlop:
I'm a bit hesitant about complexity and recipe duration. It seems a lot like titanium? 

I will load it in game I feel it's a bit hard to get everything from reading it on GitHub.

Pyritie:
yeah, I read that irl zirconium processing is almost identical to titanium, at least for "general use" zirconium. It being so similar also encourages the player to move their titanium processing line to venus too, so they can use the same setup for both

TomPlop:
Ooh I see what you mean thematically. Though using the same line for two materials shouldn't be encourage I think. Especially titanium. 

If Zirconium is one of the EBF coils in place of HSS I'm a bit hesitant to have it that easy. But well, obtaining Zircon dust can be itself a rough process.

Let me launch the game so my poor brain can see the numbers ingame.

Pyritie:
I don't think it needs to be all that complicated -- they'll already have to do a new process to set up the advanced oreproc (and zirconium makes it mandatory), and some of the other Zr materials need bromine (which is another whole line)

kubejs/startup_scripts/tfg/materials.js
```diff
@@ -553,8 +554,59 @@ const registerTFGMaterials = (event) => {
...
+		.components('1x zirconium', '4x bromine')
+		.flags(GTMaterialFlags.DISABLE_DECOMPOSITION)
+
+	event.create('tfg:zirconium_diboride')
+		.ingot()
+		.color(0xFADED2)
+		.secondaryColor(0x4FA883)
+		.iconSet('shiny')
+		.components('1x zirconium', '2x boron')
+		.blastTemp(4600, "high", GTValues.VA[GTValues.EV], 2700)
```
TomPlop: I would push it to 5000 Kelvin so the cooling requires liquid Helium as it will be the case mostly everywhere at IV

Pyritie:
I'm not on my phone rn but I remember the second argument (the "high") is
what determines the freezer gas, not the number - the number is for what
coils you need

kubejs/startup_scripts/tfg/materials.js
Pyritie: I'm not on my pc rn but I remember the second argument (the "high") is
what determines the freezer gas, not the number - the number is for what
coils you need

TomPlop: Yeah but high means the EBF recipe will use Argon which is good but the temp of coil will determine if the Freezer requires a better fluid to cool it down.

I'm pretty sure if we push the strings on .blastTemp we may be able to modify it manually. I may test that.

kubejs/startup_scripts/gtceu/materials.js
```diff
@@ -295,6 +297,11 @@ const registerGTCEuMaterialModification = (event) => {
 	GTMaterials.Plutonium239.addFlags(GENERATE_ROD, NO_ORE_PROCESSING_TAB, NO_ORE_SMELTING)
 	GTMaterials.Thorium.addFlags(NO_ORE_SMELTING)
 
+	GTMaterials.Zirconium.addFlags(GENERATE_FINE_WIRE, GENERATE_PLATE, NO_ORE_SMELTING);
+	GTMaterials.Zirconium.setProperty(PropertyKey.INGOT, new $INGOT_PROPERTY());
+	// same stats as titanium
+	GTMaterials.Zirconium.setProperty(PropertyKey.BLAST, new $BLAST_PROPERTY(1941, 1, 480, 1500, 480, 144));
```
TomPlop: GTMaterials.Zirconium.setProperty(PropertyKey.BLAST, new $BLAST_PROPERTY(4200, 'mid', GTValues.VA[GTValues.EV], 1300, GTValues.VA[GTValues.HV], 14.7*20));

I would go with that simply so it's on par with HSS-G which would be its equivalent. It would feel weird to have lower values for higher materials.

I'm tempted to add string to see if we can modify fluid in the Vacuum Freezer lol

kubejs/startup_scripts/tfg/materials.js
TomPlop: .blastTemp(5000, "high", GTValues.VA[GTValues.EV], 2700, GTValues.VA[GTValues.HV], 12.5*20)

I would go with that so it aligns with HSS-E, which is kinda its equivalent.

kubejs/server_scripts/tfg/recipes.space_ores.js
```diff
@@ -202,4 +202,60 @@ function registerTFGSpaceOres(event) {
...
+		.itemOutputs('5x #forge:dusts/zirconium_tetrachloride')
+		.duration(20 * 20)
+		.EUt(GTValues.VA[GTValues.EV])
+
+	event.recipes.gtceu.electric_blast_furnace('tfg:zirconium')
+		.itemInputs('2x #forge:dusts/magnesium', '5x #forge:dusts/zirconium_tetrachloride')
+		.itemOutputs('#forge:hot_ingots/zirconium', '6x #forge:dusts/magnesium_chloride')
+		.duration(40 * 20)
+		.blastFurnaceTemp(2141)
+		.EUt(GTValues.VA[GTValues.HV])
```
TomPlop: event.recipes.gtceu.electric_blast_furnace('tfg:zirconium')
		.itemInputs('2x #forge:dusts/magnesium', '5x #forge:dusts/zirconium_tetrachloride')
		.itemOutputs('#forge:hot_ingots/zirconium', '6x #forge:dusts/magnesium_chloride')
		.duration(55 * 20)
		.blastFurnaceTemp(4241)
		.EUt(GTValues.VA[GTValues.EV])

kubejs/startup_scripts/tfg/materials.js
TomPlop: I'm getting a crash on that one now I thought it was working I'm quite confuse.

TomPlop:
Sounds good to me now.

We really need to begin having a better repository of every materials we make on the google sheet or something.

PR #2211, opened 2025-11-12
Support Recycling

Nebby1999:
## What is the new behavior?
* This PR adds a way to recycle supports utilizing macerators or millstones

## Implementation Details
* I tried to place the recipes in places that made sense, this proved to be more difficult than i expected.

## Outcome
* Supports can now be recycled.
* Most of them should be relatively lossy, except for the steel support, which is lossless

## Additional Information
I hate javascript

* Probably none

* Yknow who it is

* Still keeping my role for a bit longer baybeeee


Redeix (approved):

Pyritie:
also sometimes I wonder if all this would be easier if we turned all the wood types into materials and made supports a tagprefix lol

PR #2216, opened 2025-11-13
add  rock/smooth tag to smooth stone

Nebby1999:
## What is the new behavior?

* Adds the specified tag to smooth stone so that it can be used to craft comparators n others

## Implementation Details

It just adds a tag

## Outcome

I can use smooth stone for comparators :D

## Additional Information

None

None

Yknow who it is


Pyritie:
you don't need to pr tiny things like this bro lol

PR #2219, opened 2025-11-14
fix railgun ammo size

TomPlop:
stack from 4 to 16


Redeix:
Let's see how many features we can get in before maven works

PR #2220, opened 2025-11-14
Add assembler and smelting recipes for firmalife ovens

SpicyNoodle5:
## What is the new behavior?
Adds assembler and furnace recipes for firmalife oven blocks

## Implementation Details
Recipes added in the same way as other assembler/furnace recipes for unfired/fired clay products

## Outcome
This allows the electric oven recipes to be fully automated, before you'd have to knap and cure a separate top oven by hand

Discord: Spicy Noodles


Pyritie:
I do wonder if people will see these jei recipes and then incorrectly assume they can't fire one until steam age

I wonder if they should have tfc heating recipes as well?

Redeix:
How do the circuits look? Feels to me like there isn't even enough circuit numbers for all of these assembler clay recipes.

SpicyNoodle5:
The circuit numbers go up to 32, which is exactly all the available circuit numbers. Could take out the fire clay recipes and add them separately tho, they don't really need to be with the regular clay recipes here. Also adding heating recipes doesn't sound bad, depends if you don't mind making the firmalife curing mechanic redundant or not

TomPlop:
> I do wonder if people will see these jei recipes and then incorrectly assume they can't fire one until steam age
> 
> I wonder if they should have tfc heating recipes as well?

maybe a massive tooltip on it

Pyritie:
> > I do wonder if people will see these jei recipes and then incorrectly assume they can't fire one until steam age
> > I wonder if they should have tfc heating recipes as well?
> 
> maybe a massive tooltip on it

nobody reads

Redeix:
> The circuit numbers go up to 32, which is exactly all the available circuit numbers. Could take out the fire clay recipes and add them separately tho, they don't really need to be with the regular clay recipes here. Also adding heating recipes doesn't sound bad, depends if you don't mind making the firmalife curing mechanic redundant or not

This sounds good to me 👍

Pyritie:
on the other hand, telling people they need to cure their IV multiblock food oven would be very funny

SpicyNoodle5:
> on the other hand, telling people they need to cure their IV multiblock food oven would be very funny

Lmao that's exactly what prompted me to make this PR, surely by the time you can head to space there's better ways of making an oven

kubejs/server_scripts/firmalife/recipes.js
```diff
@@ -1015,4 +1015,21 @@ const registerFirmaLifeRecipes = (event) => {
 			},
 			"duration": 16000
 		}).id('tfg:barrel/shosha_wheel')
+	
+	// #region Oven heating recipes
+	
+	event.recipes.tfc.heating(`firmalife:oven_top`, 1399)
+			.resultItem(`firmalife:cured_oven_top`)
+			.id(`firmalife:oven_top`)
```
Pyritie: I know we're bad at it but let's try to keep new IDs to the tfg namespace (unless they're intentionally overwriting an existing recipe)

SpicyNoodle5: Noted, updated those

Redeix (approved):

PR #2234, opened 2025-11-17
Add shapeless p2p recipes

SpicyNoodle5:
## What is the new behavior?
Adds shapeless recipes for p2p -> p2p. This is to quickly clear names from p2p tunnels similar to how you can clear filters in the crafting grid.

## Outcome
With the advanced memory card, you can't remove names from p2p tunnels completely so this acts as a convenience.

Discord: Spicy Noodles (lol)

### DIFF
--- a/kubejs/server_scripts/ae2/recipes.js
+++ b/kubejs/server_scripts/ae2/recipes.js
@@ -327,6 +327,17 @@ const registerAE2Recipes = (event) => {
 		D: '#forge:gems/fluix',
 	}).addMaterialInfo().id('tfg:crafting/me_p2p_tunnel')
 
+	event.shapeless('ae2:me_p2p_tunnel', ['ae2:me_p2p_tunnel'])
+		.id('tfg:shapeless/me_p2p_tunnel')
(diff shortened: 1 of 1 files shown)


TomPlop:
Maybe that's something that could be added as a native feature in AE2. I never encountered this issue though.

@ko-lja thought?

SpicyNoodle5:
Its only really an issue when using the advanced card as it allows you to give p2ps names but doesn't let you remove them once you have so if you're rewiring a network you can end up with lots of differently named p2ps in your inventory

TomPlop:
I'm a bit confuse because I thought you could reset them fully with the Better P2P card?

SpicyNoodle5:
Idk unless I'm missing something whenever I try to fully clear the names they just reset back to the name they had

ko-lja:
The renaming with the advanced memory card has always been a little weird, this should be fine to merge imo

TomPlop:
Not gonna lie I just realize now that it's a pull request 😂

Okay thanks let's approve it then sounds like a simple fix.

Pyritie (approved):

Redeix (approved):

PR #2282, opened 2025-11-23
Update gating of Sophisticated Backpacks upgrades

CrashAndSideburns:
## What is the new behavior?
Currently, the pickup upgrade and advanced pickup upgrade for Sophisticated Backpacks require HV and EV components, respectively. This makes them completely useless, since the magnet upgrade and the advanced magnet upgrade, which are generally better than the pickup upgrades, are available at LV and MV, respectively. This change makes the pickup upgrades both accessible at LV and moves the magnet upgrade back to MV, making the pickup upgrade into a crafting component for the magnet upgrade.

The decision to move the magnet upgrade to MV may be questionable, since it retroactively makes something harder to craft. In practice, the best use for the magnet upgrade is probably for automatically picking up ores while mining, but I haven't played enough of the pack to know if a player will still be manually excavating ore veins in MV. If most players would be using some sort of auto-miner by this point, gating the magnet upgrade to MV makes it far less useful.


Pyritie:
if you'd like to move the pickup upgrade even earlier, you could treat the create mechanical arm as a "ulv robot arm" if you'd like? I'd be fine with that

CrashAndSideburns:
I tried throwing together a new recipe which should move the pickup upgrade to ULV, and keep the magnet upgrade in LV. I tried to keep the pickup upgrade recipe similar, just using some Create components which I thought reasonably approximated the LV components. Let me know if you think that it makes sense!

Pyritie:
looks good to me!

BlueBoat29:
You can add your discord username if you would like the contributor rule

CrashAndSideburns:
I haven't got a Discord account, I'm afraid. Thanks for offering though, and thanks for the great modpack! :heart:

PR #2298, opened 2025-11-28
TFC Gurman integration

Pyritie:
Most of the recipes are by @TerraFirmaBlade, but I did some cleanup and other integration.

Not sure if all of the shaped/shapeless recipes need food processor ones, and the drinks don't have any special effects right now. Would be nice to maybe give some of them heating/cooling effects?


kubejs/server_scripts/tfg/recipes.food.js
```diff
@@ -678,7 +688,7 @@ function registerTFGFoodRecipes(event) {
 		)
 	})
 
-	processorRecipe("pizza_1_extra", 600, 16, {
+	processorRecipe("pizza_1_extra", 600, GTValues.VHA[GTValues.MV], {
```
Pyritie, line 699: gurman removes the original firmalife pizza recipe for its own pizzas, but I think keeping it as an advanced 5-nutrient food option would be fine

Redeix (approved):
Looks good to me. But I thought we wanted to avoid more complex foods in the meal bag?

Pyritie:
sure, I can remove a bunch of them, though a lot of those felt more like condiments or sauces than something you'd eat by yourself 🤔

Redeix:
> sure, I can remove a bunch of them, though a lot of those felt more like condiments or sauces than something you'd eat by yourself 🤔

No, Im not against it. I thought it was your idea since you removed the banana split i added T.T

PR #2306, opened 2025-11-29
changed coloured blast steel recipes to be mv

applenper:
coloured steel dust to ingot recipes are now MV tier instead of HV tier in the electric blast furnace

app.le on discord

### DIFF
--- a/kubejs/server_scripts/gregtech/recipes.tfcmetals.js
+++ b/kubejs/server_scripts/gregtech/recipes.tfcmetals.js
@@ -34,6 +34,45 @@ function registerGTCEuTFCMetalsRecipes(event) {
 		.duration(700)
 		.EUt(GTValues.VA[GTValues.LV])
 
+	event.remove({id:'gtceu:electric_blast_furnace/blast_red_steel_gas'})
+	event.remove({id:'gtceu:electric_blast_furnace/blast_red_steel'})
+	event.remove({id:'gtceu:electric_blast_furnace/blast_blue_steel_gas'})
+	event.remove({id:'gtceu:electric_blast_furnace/blast_blue_steel'})
+
+	event.recipes.gtceu.electric_blast_furnace('blast_red_steel_gas_mv')
+		.itemInputs('gtceu:red_steel_dust')
+		.itemOutputs('tfc:metal/ingot/red_steel')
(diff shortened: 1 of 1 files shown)


SpicyNoodle5:
Blue and red steel ingots already have LV dust -> ingot recipes in the arc furnace, perhaps the EBF recipes should just be removed entirely?
image

applenper:
> Blue and red steel ingots already have LV dust -> ingot recipes in the arc furnace, perhaps the EBF recipes should just be removed entirely? image

perhaps, though the option is nice to have as the EBF recipe uses nitrogen instead and benefits from perfect overclocking

Pyritie:
yeah, the overclocking is handy

Pyritie (approved):

TomPlop:
It feels really weird to have both though

Shouldn't we make both runs in MV EBF with Blacksteel too and scrap the arc furnace recipe? The arc furnace recipes doesn't feel rewarding because of instant solidification with the crucible. So it's like in a way a trap (though it's good for glass)

We should uniform all of them into (Black, Blue and Red) into MV EBF and remove the Arc furnace recipes.

Pyritie:
mm, the arc furnace recipes were one of my really early contributions in 0.7, mostly just cuz of all the red steel you'd use in LV, but that was before everyone now having a fully automated crucible into casting table setup. I originally made it arc furnace just so it was LV locked, instead of the regular furnace which the steam one could do too

TomPlop:
> mm, the arc furnace recipes were one of my really early contributions in 0.7, mostly just cuz of all the red steel you'd use in LV, but that was before everyone now having a fully automated crucible into casting table setup. I originally made it arc furnace just so it was LV locked, instead of the regular furnace which the steam one could do too

Yeah that's the fact that everyone uses Create and Crucible that make me think we should backtrack on the Arc furnace and just fully support EBF at MV (or even LV?). We could also put a really low temperature to make it even easier to POC.

I get you because even when I was making the quests for LV I put a lot of emphasis on the Arc furnace recipes but since then it feels this emphasis is a bit out of scope because of the Create/Crucible thing. Maybe we should fully revert the Arc Furnace now and just support an easier EBF recipe.

Pyritie:
yeah, the arc furnace thing does feel redundant now

PR #2318, opened 2025-12-03
Tallow and fish oil changes for 0.11.9

applenper:
rather than replacing fish oil with tallowate altogether, we can just use tallow to make biodiesel directly, which is more realistic and lets us preserve fish oil

also making tallow part of firmalife:oils and making it a boiler fuel as well so it's in-line with other oils

removed the tallow -> fish oil recipe because they're functionally identical now

i also buffed the output of getting tallow and fish oil from extractors from 200mb to 400mb to compensate for this. subject to balancing ofc

p.s. since blubber is being replaced by animal fat, currently the only way to get fish oil now is by extracting fish. maybe we want to add a new way of getting fish oil in the future. (probably make blubber and animal fat separate items eventually, but it doesn't really matter right now since again, fish oil and tallow are pretty much identical now)

app.le on discord

### DIFF
--- a/kubejs/assets/gtceu/lang/en_us.json
+++ b/kubejs/assets/gtceu/lang/en_us.json
@@ -132,7 +132,7 @@
   "material.gtceu.damascus_steel": "Swine-steel",
   "material.gtceu.air": "Earth Air",
   "material.gtceu.liquid_air": "Liquid Earth Air",
-  "material.gtceu.fish_oil": "Tallowate",
+  "material.gtceu.fish_oil": "Fish Oil",
   "material.gtceu.cooperite": "Cooperite",
   "material.gtceu.armalcolite": "Armalcolite",
   "material.gtceu.desh": "Desh",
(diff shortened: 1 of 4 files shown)


kubejs/assets/gtceu/lang/en_us.json
```diff
@@ -132,7 +132,7 @@
   "material.gtceu.damascus_steel": "Swine-steel",
   "material.gtceu.air": "Earth Air",
   "material.gtceu.liquid_air": "Liquid Earth Air",
-  "material.gtceu.fish_oil": "Tallowate",
+  "material.gtceu.fish_oil": "Fish Oil",
```
TomPlop, line 135: Sorry but could you do the change in the Tool Modern repo please that's where we have our actual langfile :p

TomPlop (approved):
Let's go

Pyritie (approved):

PR #2328, opened 2025-12-05
Replace bronze drums & crates in recipes with new tags containing all 3 bronze drums & crates

### LINKED ISSUE #2327: [🪶]: Basic Super Chests should use the alternative bronze crates in the crafting recipe

- [x] No, it was not.

### Version modpack used

0.11.8

### Related problem

Currently the recipe for the Basic/ULV Super chest only accepts the normal Tin Bronze Recipe.

### Your solution

The Bismuth Bronze and Black Bronze crates should also be allowed as substitutes, for the sake of consistency as these materials are usually interchangeable in other recipes.

### Additional information

Image

Jeuvke:
## What is the new behavior?
Added bronze, bismuth bronze, and black bronze drums and crates to their own tags (tfg:any_bronze_drum and tfg:any_bronze_crate). Replaced all bronze drums and crates in existing recipes with the new tags.
Fixes #2327.

## Implementation Details
This PR uses event.replaceInput with recipe input filtering, but without recipe ID filtering. Currently there are no unintended consequences from this, as GT's recycling recipes are exempted from replaceInput.


kubejs/server_scripts/tfg/recipes.miscellaneous.js
```diff
@@ -95,7 +95,11 @@ function registerTFGMiscellaneousRecipes(event) {
 		.itemOutputs('tfg:wireless_card')
 		.duration(300)
 		.EUt(GTValues.VA[GTValues.EV])
-	
+
+	// Replace bronze drums & crates with a tag containing the 3 different bronzes
+	event.replaceInput({}, 'gtceu:bronze_crate', '#tfg:any_bronze_crate')
```
Pyritie: would prefer if you at least did `mod: 'gtceu'` for the filter instead of doing literally everything, should help a little with performance

Pyritie (changes requested):

kubejs/server_scripts/tfg/recipes.miscellaneous.js
Jeuvke: I don't think that works, as the super tank/chest recipes are tfg recipes.

Pyritie: `mod: 'tfg'` then? 🤔 this just feels like it'd be expensive

Jeuvke: In the log it says it took 0.002s to load the script, so I believe it's ok?

Pyritie: ah maybe kjs is doing something clever then

Pyritie (dismissed):

Jeuvke:
I'm really unhappy with how I'm doing this, but I can't yet figure out a way to do this more cleanly.

Pyritie:
given the limited set of items, would it be possible to just loop over them and replace the inputs or create alternative recipes that way?

Jeuvke:
I'll just add some input filtering and call it a day. There doesn't appear to be a way to do it the "proper" way, by setting the GT DRUM and CRATE tags to the new any_bronze_drum and any_bronze_crate item tags, as the item tags get registered after the recipes.

Issue #2347, opened 2025-12-08
[💥]: `TypeError: Cannot read property "CREATE_BASE" from undefined` in Kube.js


mgsmyrk:
### Was this crash report before?

- [x] No, it was not.

### Version modpack used

0.11.9

### Environment

Startup

### Side Crashed

Client

### New Worlds

Unsure

### Please provide a link to your latest.log, if relevant

https://mclo.gs/S8gtrqN

### Steps to Reproduce

On Linux (Ubuntu 24.04), with Prism Launcher 9.1.

This crash happens on startup with both a new Minecraft instance and upgrading an already-working 0.11.8 instance.

Game crashes on startup before displaying the main menu.

mgsmyrk:
It looks like the issue might be to do with the order the startup scripts are loaded. 

CREATE_BASE is defined in constants.js. Changing main_startup_script.js:73 to use the 'create:base' constant directly works around the problem.

Pyritie:
would you be able to send a pull request for that? I don't have this issue on my own PC (windows) so I'm not able to test it properly

Nekitx:
faced the same issue on arch linux also in prism

mgsmyrk:
Will do in a few hours

mgsmyrk:
A temporary workaround for anyone experiencing this while waiting for a proper fix to be released:

* In your minecraft folder, open `kubejs/startup_scripts/main_startup_script.js` in a text editor. 
* On Line 73 (top of the second last block of text), replace `global.TAB_IDS.CREATE_BASE` with `'create:base'`.

So the line should be: 
`StartupEvents.modifyCreativeTab('create:base', event => {`

This should be enough to get the game running again.

PR #2351, opened 2025-12-08
2347:  Moved the tab name constant from constants.js to main_startup_script.js.

mgsmyrk:
## What is the new behavior?

https://github.com/TerraFirmaGreg-Team/Modpack-Modern/issues/2347

## Implementation Details

Is seems the startup_script ordering is unstable, causing the main script to load before constants.ts on Linux. 

I have moved the string constant from constants.ts, and hard coded it in the only place it's used.

## Outcome

Fixes: #2347 

None

msmyrk6186


Redeix:
This is not a proper solution. Ideally we should just create a new category in main scripts to load all constants before everything else. I don't really have time to do that right now but if you or anyone else want to do that I will approve the merge.

mgsmyrk:
I'll have another crack at this in a few hours. (Sorry, new to MC mods and KubeJS).

Redeix:
All good. This is something that's caused Linux problems for a while now and I've been meaning to change it. If you can't get around to it I'll fix it when I get off work tonight 👍

Hydroyeti1:
Is it silly to just up the priority of the constants script?

mgsmyrk:
I couldn't work out how to add a new category in a way that would have deterministic order. Not a fan of setting the priority (feels flaky).

It turns out KubeJS has a "requires" header (https://kubejs.com/wiki/tips/headers), which lets one script force another to load first. I've made main_startup_script depend on constants. What do you think @Redeix ?

Redeix:
I agree that priority order is not ideal. But the "requires" header only works for requiring mods to be loaded. Not specific files

mgsmyrk:
Hmm. You're right. I just tried reversing the dependency and it doesn't break it. Strangely, my change "fixes" the crash though (I even tried deleting and re-creating each file when testing to make sure I hadn't fixed it with file creation ordering). 

Okay, so the correct fix is to move constants.js to common/constants.js, then make main_startup_script depend on "common"?

Pyritie:
I think it would be easier to just merge the files together

Redeix:
I'm not against just merging all the constants together under one function and then just leading that in main. Otherwise we will need to declare all other constants in there own seperate functions that get called in main in sequence

Pyritie:
not all the constants, just this one for this case, since it's a bit of an exception

Redeix:
I still want all constants loaded before the rest of the startup scripts so we never have to deal with this again. It's really a simple fix.

mgsmyrk:
In this specific case, this is the only constant needed by main_startup_script, and the constant is not used anywhere else. I get your point that a proper solution would be good moving forward though.

@Redeix Can you point me to documentation for what you mean by "creating a new category"? I'm happy to make the change, but I'm lacking domain knowledge here.

Redeix:
I would say to just let me take care of it. I need to test a lot of stuff to find the proper solution. Im essentially arguing to do what we already do for all recipe scripts. Which is to place everything into a function that then gets called in the main server script before everything else. It can't be done exactly the same since they aren't events.

mgsmyrk:
No worries - I'll just leave you with this last version of mine, which seems a fairly simple solution. Tested to fix the bug (also tested that reversing the dependency breaks it).

mgsmyrk:
By the way @Redeix , if I'm understanding your proposed fix correctly (to call a function to either set the global or return the config object), then I don't think that will work unless you can change things to call `StartupEvents.modifyCreativeTab` from **within** some other event handler.

You're likely to run into an issue where the function doesn't get declared until after modifyCreativeTab handler gets registered. The reason this works with the recipes configs is those functions are being called from within an event handler so it doesn't matter what order the code is loaded. The functions get called by Kube's event manager after all the JS has been loaded (so all the functions defined).

This case is different because the constant is needed at the time of registration, not event handler execution.

Redeix:
Okay so I got home and actually looked at the issue here and tbh there was no reason to even declare the creative tab scripts in main to begin with. So I got rid of the constant and registry and just kept it by itself. I did go through and update all the priorites just to make sure we dont run into further issues. What I wanted to do would have worked but it would have been too much effort due to the way that global scope works in JS. (why doesnt rhino just support classes?) so oh well.
Im going to close this PR, thank you for looking into this for us. If you wouldnt mind, could you update your fork and double check that latest works on Linux? I dont have a machine to test it on.

mgsmyrk:
@Redeix I just ran my instance with your latest change, and it boots fine on Linux.

Pyritie:
great, thank you for all your testing!

PR #2356, opened 2025-12-09
Add Screwdriver to the drums' patchouli page

jmecn:
## What is the new behavior?
Add Screwdriver to the drums' patchouli page

## Implementation Details
It was remove since 0.10.15, but add back in 0.11.18.

https://github.com/TerraFirmaGreg-Team/Modpack-Modern/blob/0d75405aee6a00d522e1e175b4da3ab51b90265c/kubejs/assets/tfg/lang/en_us.json#L2853


Pyritie:
ah good catch, I forgot that it was in the field guide too

PR #2384, opened 2025-12-11
Fix quest MV quad pipe accepted items

Mqrius:
Accept any quadruple or nonuple pipe based on item tags, instead of one of three specific items. This puts it in line with the actual quest text.
image
image


Jeuvke:
Finally someone fixes this 🙏

Redeix (approved):

Pyritie (approved):

PR #2387, opened 2025-12-12
Add circuit to Silicon Tetrachloride

### LINKED ISSUE #2383: [🐛]: /Polydimethylsiloxane dust recipe making silicon tetrachloride

- [x] No, it was not.

### Version modpack used

0.11.10

### Environment

singleplayer

### New Worlds

Unsure

N/A

### Expected Behavior

enter ingredients for PDMS dust, as seen here, on circuit #2

Image

### Actual Behavior

consumed silicon and chlorine to make silicon tetrachloride instead, as STC is non-circuit dependent.

### Steps to Reproduce

attempt to make PDMS dust.

### Additional Information

obviously, it's just because STC needs to be circuit dependent, please fix it.

Mqrius:
Fixes #2383, deconflicts with Polydimethylsiloxane (and various other recipes!)

Pretty sure this doesn't introduce any new conflicts, since there's no chemreactor recipes that have circuit one and only silicon, only chlorine, or both.

### DIFF
--- a/kubejs/server_scripts/tfg/ores_and_materials/recipes.zirconium.js
+++ b/kubejs/server_scripts/tfg/ores_and_materials/recipes.zirconium.js
@@ -9,6 +9,7 @@ function registerTFGZirconiumRecipes(event) {
 		.inputFluids('gtceu:chlorine 4000')
 		.outputFluids('tfg:silicon_tetrachloride 1000')
 		.duration(20 * 20)
+		.circuit(1)
 		.EUt(GTValues.VA[GTValues.MV])
 
 	event.recipes.gtceu.chemical_reactor('tfg:silicon_tetrachloride_decomp')


TomPlop:
Be very wary when changing circuits. If it's a recipe that's already implemented it has to be noted has critical changes and if not implemented yet then no one cares.

Mqrius:
@Tom This is a recipe introduced in the zirconium line right? So it should be essentially new. It really does need a circuit though cos there's a whole lot of other recipes that have silicon and chlorine as some of the ingredients. Or am I misunderstanding what you're saying?

Edit: Okay not _that_ many, there's just the kaolinite powders and the PDMS, but still
image
image
image

TomPlop:
yeah 

> @tom This is a recipe introduced in the zirconium line right? So it should be essentially new. It really does need a circuit though cos there's a whole lot of other recipes that have silicon and chlorine as some of the ingredients. Or am I misunderstanding what you're saying?
> 
> Edit: Okay not _that_ many, there's just the kaolinite powders and the PDMS, but still image image image

Perfect was just noticing in case

TomPlop (approved):

Pyritie:
maybe we should add a circuit to the kaolin one too

Mqrius:
> maybe we should add a circuit to the kaolin one too

Since it contains many different ingredients it's actually less necessary to give it a circuit I'm thinking. It's also an existing recipe already so as Tom said, giving it a circuit could break existing builds. Personally I'd leave it as is unless you think it should be changed?

Pyritie:
I think tom is more saying that about gregtech built-ins. The kaolin recipes are one of our own, so they should be fine, though like he says we do need to put that at the top of the changelog.

PR #2397, opened 2025-12-12
Add circuit to carbon monoxide from dioxide recipe

SpicyNoodle5:
## What is the new behavior?
Adds circuit 1 to the carbon monoxide from dioxide recipe. This is to help with a recipe "conflict" in the tungsten line if you use the same chemical reactor for both carbon dioxide and sodium bicarbonate (carbon + carbon dioxide will react first to make useless carbon monoxide rather than using carbon + oxygen to make carbon dioxide)

## Implementation Details
Adds circuit 1, this matches other recipes for carbon monoxide which also have circuit 1

### DIFF
--- a/kubejs/startup_scripts/gtceu/constants.js
+++ b/kubejs/startup_scripts/gtceu/constants.js
@@ -387,6 +387,9 @@ global.ADD_CIRCUIT = /** @type {const} */ ([
 
     { recipeId: "gtceu:chemical_reactor/sodium_bicarbonate_from_salt", circuitNumber: 2 },
     { recipeId: "gtceu:large_chemical_reactor/sodium_bicarbonate_from_salt", circuitNumber: 2 },
+
+    { recipeId: "gtceu:chemical_reactor/carbon_monoxide_from_dioxide", circuitNumber: 1 },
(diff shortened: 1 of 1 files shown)


TomPlop:
Won't circuit one conflict then with Formic Acid? In this kind of case I would prefer we don't change circuit and the player use colored hatches on their LCR

SpicyNoodle5:
This recipe is carbon + carbon dioxide, formic acid is carbon dioxide + water, tho if it doesn't have a circuit to begin with won't it be a "conflict" anyway? Or it could go on circuit 0 or 5 which don't share any other recipes with either carbon or carbon dioxide. Curious what you mean about coloured hatches, didn't know that actually did anything

Pyritie:
we really should make a quest for the colored hatches

SpicyNoodle5:
Huh the colour thing just works lol

TomPlop:
I thought I made a quest for color hatches. Anyway we can close this PR. Overall I would prefer if the only modifications we do on circuits is in case a recipe is impossible to do because it's conflicting.

Pyritie:
would be good to copy that quest somewhere more noticeable then - maybe right above the tungsten line!

PR #2416, opened 2025-12-14
Add btx

TomPlop:
### Breaking Changes
- The recipe duration of the Raw Aromatic Mix in the Coal Liquefaction Tower were change to align with Pyrolyse Oven. You now need 2 Pyrolyse Oven for 1 Liquefaction Tower If they have the same energy tier and Coils.

## Changes
- The new BTX Fuel is now in game and unlocked at EV
- The Coal Liquefaction Tower was rename to Liquefaction Tower
- Liquefaction Tower now use the same logic as the Pyrolyse Oven and upgrading coils reduce recipe duration

### DIFF
--- a/kubejs/server_scripts/tfg/powergen/recipes.early_gas.js
+++ b/kubejs/server_scripts/tfg/powergen/recipes.early_gas.js
@@ -8,41 +8,29 @@ function registerTFGEarlyGasRecipes(event) {
 
 	//#region Syngas
 
-	event.custom({
-		type: 'vintageimprovements:vacuumizing',
-		ingredients: [{ item: 'minecraft:charcoal' }, { fluid: 'gtceu:creosote', amount: 250 }],
-		results: [{ fluid: 'tfg:syngas', amount: 100 }],
-		secondaryFluidOutput: 0,
-		heatRequirement: "heated",
-		processingTime: 4000
-	}).id('tfg:vi/vacuumizing/syngas_from_charcoal')
+	event.recipes.vintageimprovements.vacuumizing(Fluid.of('tfg:syngas', 100), ['minecraft:charcoal', Fluid.of('gtceu:creosote', 250)])
+		.secondaryFluidOutput(0)
+		.processingTime(4000)
+		.heated()
+		.id('tfg:vi/vacuumizing/syngas_from_charcoal')
 
-	event.custom({
-		type: 'vintageimprovements:vacuumizing',
-		ingredients: [{ item: 'minecraft:coal' }, { fluid: 'gtceu:creosote', amount: 250 }],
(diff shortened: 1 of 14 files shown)


kubejs/server_scripts/tfg/powergen/recipes.early_gas.js
```diff
@@ -240,10 +238,143 @@ function registerTFGEarlyGasRecipes(event) {
...
+		.outputFluids(Fluid.of('gtceu:methanol', 4000), Fluid.of('minecraft:water', 2000))
+		.duration(20*12)
+		.EUt(GTValues.VA[GTValues.MV])
+
+	//#region BTX Fuel
+
+	// ZSM - 5
+
+	event.recipes.gtceu.chemical_reactor('tfg:sodium_silicate')
+		.itemInputs(Item.of('#forge:dusts/nether_quartz'))
```
Pyritie, line 252: do we want to allow regular quartzite as well?

kubejs/startup_scripts/tfg/items.js
```diff
@@ -68,4 +68,18 @@ const registerTFGItems = (event) => {
 	event.create('gtceu:refined_rose_quartz_ore').texture('tfg:item/deprecated')
 	event.create('gtceu:rose_quartz_lens').texture('tfg:item/deprecated')
 	event.create('gtceu:rose_quartz_plate').texture('tfg:item/deprecated')
+
+	// Add item for BTX Fuel
```
Pyritie, line 72: let's move these elsewhere so items.js doesn't become a mess again

Pyritie (changes requested):

kubejs/startup_scripts/tfg/items.js
TomPlop: I'm unsure about making a file for 3 items though

Pyritie: maybe we could rename the nuclear folder to "energy" or "power", to keep all of our power stuff together? I'm sure fusion will involve new stuff too

kubejs/server_scripts/tfg/powergen/recipes.early_gas.js
TomPlop: I prefer nether quartz only for balancing reasons (quartzite has a lot of other usage)

Pyritie: I suppose

kubejs/startup_scripts/tfg/items.js
TomPlop: we could but then we could move the EV progression stuffs to Mars or something

We should merge the PR before in any case

Issue #2374, opened 2025-12-10
[🐛]: Drills (the tool one) durability drains faster than expected.


Pawel215:
### Was this bug reported before?

- [x] No, it was not.

### Version modpack used

0.11.9

### Environment

Singleplayer

### New Worlds

Yes

### Please provide a link to your latest.log, if relevant.

not

### Expected Behavior

In version 0.10.0 drills durability depleted much slower aproximetly
tested with all tungsteen carbide and blue steel drills (from lv to iv). 
tested in 0.10.0 and 0.11.9 

### Actual Behavior

durability consumed was equal to blocks mined so drill heads  break faster
it didnt happen in 0.10.0, but happend in 0.11.9

mined  9x9x9 so 729 43007-42192= 815. actually more than predicted 
42192-41398 = 794. didnt test 1x1x1, since it doesnt mather in great scheme of things.

<img width="1597" height="1182" alt="Image" src="https://github.com/user-attachments/assets/87540051-2d73-4c0c-abad-86cbef24441b" />
forgot to take a second photo,

### Steps to Reproduce

Take a drill mine like 10 blocks and have 10 durability gone

### Additional Information

0.10.0 because it was my last version before break. bug was found (i did notice it but didnt think of it by DosTanense on dsc xd.
guessing that it might be on the gregtech side so not possible to change or maybe intended? please no xd

Pyritie:
can you test in 0.11.10 as well? if the bug is still there, please report it to gregtech instead, it'll be their issue to fix

Pawel215:
nwm i dindt even know it released, updating now. or maybe i did id remember

Pawel215:
<img width="1020" height="377" alt="Image" src="https://github.com/user-attachments/assets/9d9ce119-dec5-4139-9a5c-c86bf0f0e678" />
mined 10 blocks 12 durability lost  drill started at 43008
here 10 blocks 11 durability 

<img width="999" height="790" alt="Image" src="https://github.com/user-attachments/assets/dbe4574e-1dd4-49d5-9322-b1396300073b" /> 
32 mined total lost 31
guessing there a chance to loss +1 or nothing going to test 9^3 next

Pawel215:
43007-42204 = 803 9^3= 729
i mean as long it was intened to loss 1 durability per block it doesnt mather 
0.11.10
[...]

PR #2419, opened 2025-12-14
Fix drills durability (fix sharp tools BlockEvent filter)

Mqrius:
This fixes https://github.com/TerraFirmaGreg-Team/Modpack-Modern/issues/2374

**Implementation details:**
`BlockEvents.broken('tfc:mineable_with_sharp_tool', event => {`

This is trying to filter by block tag, but you can't register to events by tag, only by ID. Since nothing matches, kubejs registers this for every block break event. (I tried the tag with # but that also doesn't work)

**Notes**
I'm not sure why kubejs doesn't just silently or loudly fail to register for events when the ID doesn't match (I'm asking on their discord) but the fix will work either way


Pyritie:
thanks a lot for tracking this down!

PR #2424, opened 2025-12-14
New Active Transformer

BlueBoat29:
like the recipes and stuff


BlueBoat29:
adding the .liquid() flag automatically adds an abs recipe

Pyritie:
oh fr

PR #2442, opened 2025-12-16
Remove localization from loading screen

Mqrius:
Localization files are only loaded after all mods are initialized. The FancyMenu loading screen is loaded _before_ all mods.
image

Currently that leads to the string "tfg.gui.menu.field _guides_ splash_text" being shown for most of the loading screen, and then near the end it briefly actually shows the correct localization, right before it disappears again to do the main menu.

This doesn't seem worth the tradeoff, it looks quite broken. imo let's revert to this being plain English.

image

### DIFF
--- a/config/fancymenu/customization/gui_loading_screen.txt
+++ b/config/fancymenu/customization/gui_loading_screen.txt
@@ -91,7 +91,7 @@ element {
 
 element {
   interactable = false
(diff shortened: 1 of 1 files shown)


Redeix:
I'm going to check first if I can cache the strings. Since I can cache other assets

PR #2454, opened 2025-12-17
New New belt textures!

Epicificator:
## What is the new behavior?

New belt block and item textures.

## Implementation Details

possibly find a way to fix the texture being slower than the items on diagonal belts.

## Additional Information

image

left 3 belt textures are new

username: "jigabit42" nickname; "JigaBit"


Epicificator:
personally i dont see a problem with the color of either of them, maybe SBR material can be adjusted to have a different ingot shape and have a slightly different color, but changing the color by a lot doesn't seem necessary.

Pyritie:
alright you win lol

PR #2491, opened 2025-12-21
Prevent placement of alcohol buckets

bigtho80:
## What is the new behavior?

This PR prevents alcohol buckets from being placed in the world in order to eliminate infinite drinking exploits. Alcohol can still be consumed or used through right-click interactions on machines, preserving its intended gameplay value and trade-offs.

## Implementation Details

* Only fluid containers holding `tfc:alcohols` or `tfcagedalcohols` are affected.
* World placement for these fluids is cancelled.
* Right-click interactions on machines remain unchanged.

## Outcome

Alcohol can no longer be placed as a world fluid, restoring meaningful cost to alcohol production while keeping all processing setups functional.

Mods or setups relying on placing `tfc:alcohols` or `tfcagedalcohols` in the world will no longer work as before. Machine interactions are unaffected.

**Discord nickname:** bigtho80


kubejs/server_scripts/tfc/cancelalcoholbucket.js
```diff
@@ -0,0 +1,7 @@
+ItemEvents.rightClicked(event => {
+    const id = event.item.id
+
+    if (id.startsWith('tfc:bucket/') || id.startsWith('tfcagedalcohol:bucket/')) {
```
Pyritie: this also cancels stuff like sea water and spring water, right? those should still be placeable

bigtho80: I changed it to use an item tag instead, and I renamed it to a more general name in case we need to prevent placing other liquids later.

PR #2508, opened 2025-12-23
add checkmark to cooked meat preservation quest for meats that are for some reason not included

thederpysockdude123:
adds a checkmark to complete the Meat Preservation: Cooking quest as fish are for some reason not included in the filter

an alternative solution could be to add oven baked fish in separately, but that sounds like a pain for very little gain.

Federation President Bravo


thederpysockdude123:
the check mark task was not marked as optional, thus it does not autocomplete, but I went ahead and removed the optional tag from the oven task anyways 👍

Pyritie (approved):
somehow I completely missed that

thederpysockdude123:
it's fine lol

PR #2514, opened 2025-12-24
Remove grappling hook warnings from quests

Mqrius:
When this PR is released, the quest warnings are no longer needed:
https://github.com/TerraFirmaGreg-Team/Core-Modern/pull/226


Pyritie (approved):
I'll merge this after the other guy finishes their quest fixes

PR #2610, opened 2026-01-04
Added Missing recipes for Steam n Rails items

thederpysockdude123:
adds recipes for the portable fuel interface, fuel tank, and wooden headstocks (via the wooden screwlink headstock which can be turned into the other wooden headstocks via the radial menu).

also changes the smokestack recipes to use cast iron rather than black steel, as they are purely decorative.

strangely, only creosote currently seems to be insertable into the fuel tank, so it will likely need to be tweaked further.

Federation President Bravo


Pyritie:
good to know that at least one fuel works! I suspect lava might also work, since it's another "default burnable" fuel?

Pyritie (approved):

PR #2620, opened 2026-01-04
Add ModernUI

woopSmile:
Added ModernUI + configs (by default keep minecraft fonts)
Removed mods that have similar features to ModernUI: blur and FPS Reducer
Changed immediately fast configs for modernUI compatibility
Screenshot 2026-01-04 180938
Screenshot 2026-01-04 181014
Screenshot 2026-01-04 181102


Pyritie:
should check if this works with moldraw

Redeix:
I'm going to add this to my single layer world and just play around for a few hours just to see how I feel about it. I like sticking to the vanilla style sometimes. But I do like a lot of the stuff I see from the mod so I just need to check it out.

Redeix (approved):
Seems good. Works with moldraw

PR #2632, opened 2026-01-05
Bugfix

Inceitious:
## What is the new behavior?
Duplicate bars

## Implementation Details
NIL

## Outcome
Removed duplicate bars

## Additional Information
Ignore all the other changed files except kubejs/server_scripts/create/recipes.js. Not sure why others are there...

NIL

Inceit

### DIFF
--- a/kubejs/server_scripts/create/recipes.js
+++ b/kubejs/server_scripts/create/recipes.js
@@ -2295,7 +2295,6 @@ const registerCreateRecipes = (event) => {
 	event.stonecutting('2x create:andesite_table_cloth', '#forge:ingots/tin_alloy')
 	event.stonecutting('2x create:andesite_scaffolding', '#forge:ingots/tin_alloy')
 	event.stonecutting('2x create:andesite_ladder', '#forge:ingots/tin_alloy')
-	event.stonecutting('2x create:andesite_bars', '#forge:ingots/tin_alloy')
 
 	// Industrial Iron stuff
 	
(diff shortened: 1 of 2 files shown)


Pyritie:
what do you mean by "duplicate bars"?

and you'll need to remove the other files from this PR before we're able to accept it. I suggest reading [this](https://github.com/TerraFirmaGreg-Team/Modpack-Modern/blob/dev/.github/CONTRIBUTING.md) on how to set up a dev environment. I have a repo elsewhere on my pc, and then symlink my kubejs folder across to my minecraft instance, so all the config files stay isolated.

Inceitious:
> what do you mean by "duplicate bars"?
> 
> and you'll need to remove the other files from this PR before we're able to accept it. I suggest reading [this](https://github.com/TerraFirmaGreg-Team/Modpack-Modern/blob/dev/.github/CONTRIBUTING.md) on how to set up a dev environment. I have a repo elsewhere on my pc, and then symlink my kubejs folder across to my minecraft instance, so all the config files stay isolated.

The stonecutter recipe shows duplicate metal bars from create, so this PR is supposed to fix that.

I see, so if I am correct to interpret, when you want to test the feature out, you duplicate the kubejs folder to your minecraft instance, so it does not carry over the config files. Noted, thanks for the advice.

Pyritie:
> I see, so if I am correct to interpret, when you want to test the feature out, you duplicate the kubejs folder to your minecraft instance, so it does not carry over the config files. Noted, thanks for the advice.

No, a [symlink](https://en.wikipedia.org/wiki/Symbolic_link) is when you tell your PC to think a folder is somewhere else. So the minecraft instance just "sees" the kubejs folder on my repo, without me having to copy anything.

PR #2660, opened 2026-01-08
moved cut and copy paste gadgets to EV

thederpysockdude123:
changed recipes of cut and copy paste builder's gadgets to be craftable at EV soon after getting to Mars

Screenshot 2026-01-07 170508
Screenshot 2026-01-07 170521

also added an alternative recipe to the cut paste gadget using 2 IV circuits instead of a kinetic core

Federation President Bravo


Redeix (approved):
I'm okay with this but we have to remember that we actually need unlockables later in the game lol. It can't just be gregtech

PR #2661, opened 2026-01-08
Give Mars tree leaves distance property

Mqrius:
Adds TFG leaf blocks for the mars trees that look and act like their original mushroom blocks, but have the leaf distance property. This way they work correctly with Falling Trees.

The leaf blocks drop their original unmodified blocks through the loot tables. There should be no way to acquire the actual leaf blocks in survival.

Alphacene is not changed at this time because that feature is directly invoked from the species mod, so it would be harder to change.

Also changed strophar vertical support to purple to match the color of the item.

PR depends on https://github.com/TerraFirmaGreg-Team/Core-Modern/pull/236 for the solid leaves tag.

### DIFF
--- a/kubejs/startup_scripts/tfg/worldgen/blocks.trees.js
+++ b/kubejs/startup_scripts/tfg/worldgen/blocks.trees.js
@@ -123,7 +123,7 @@ function registerTFGWorldGenTrees(event) {
 	// Mars leaves
 
 	event.create('tfg:crimson_wart_block', 'tfc:leaves')
-		.soundType('azalea_leaves')
+		.soundType('shroomlight')
 		.tagBlock('minecraft:mineable/hoe')
 		.tagBlock('tfg:do_not_destroy_in_space')
 		.tagBlock('tfg:solid_leaves')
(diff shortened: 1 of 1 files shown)


Redeix (approved):

kubejs/startup_scripts/tfg/worldgen/blocks.trees.js
```diff
@@ -119,4 +119,50 @@ function registerTFGWorldGenTrees(event) {
 			stripped.tagBoth('c:hidden_from_recipe_viewers')
 			stripped.soundType('wood')
 		})
+
+	// Mars leaves
+
+	event.create('tfg:crimson_wart_block', 'tfc:leaves')
+		.soundType('azalea_leaves')
```
Pyritie: I think this sound is wrong?

Redeix: I didn't realize they even had a specific sound

Mqrius: Made them shroomlight. That makes them slightly different from the blocks you can place yourself, since they're different blocks, but it's unlikely anyone will really notice. (Original blocks are either wart blocks, or ad astra's blocks which don't seem to have sound defined so maybe get "stone")

Redeix (approved):

PR #2673, opened 2026-01-08
Fix dried fruit recipe

### LINKED ISSUE #2664: [🐛]: Can't create Yeast Starter with dried fruit

- [x] No, it was not.

### Version modpack used

0.11.16

### Environment

Singleplayer

### New Worlds

Unsure

-

### Expected Behavior

I should be able to produce Yeast Starter by putting dried fruit with 100mB of Water in a Sealed Barrel, as explained in the Field Guide (I also tried 100mB increments up to 1000mB but nothing happened).

(Note that the Field Guide doesn't explain the required ratio and the EMI recipe looks broken so this is just guesswork.)

### Actual Behavior

Nothing happens:

Image

Additionally, the recipe in EMI also looks broken, as it doesn't include `tfc:foods/fruits` as an ingredient:

Image

### Steps to Reproduce

1. Dry fruit on a drying mat (I used Melon Slice)
2. Place in a barrel with 100mB of Water
3. Seal the barrel
4. Nothing happens, recipe is not recognized

### Additional Information

_No response_

Mqrius:
Fixes #2664

### DIFF
--- a/kubejs/server_scripts/firmalife/tags.js
+++ b/kubejs/server_scripts/firmalife/tags.js
@@ -18,12 +18,10 @@ const registerFirmaLifeItemTags = (event) => {
     event.removeAllTagsFrom("/tfc:ore/[^*]+/[^*]+/")
 
     // Make our own "dried fruit" tag so we can display something in EMI - used for yeast starter
-    const fruitArray = Ingredient.of('#tfc:foods/fruits').itemIds.toArray().map(String)
-    fruitArray.forEach(fruit => {
-        if (fruit !== 'betterend:shadow_berry_cooked' && fruit !== 'minecraft:popped_chorus_fruit') {
-            event.add('tfg:dried_fruit', fruit)
-        }
-    })
+    global.FOOD_FRUIT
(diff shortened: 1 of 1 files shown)


Pyritie (approved):
man I thought I tried that before and it didn't work, it must've been something else I was doing lol

Mqrius:
Actually hold on, I tried with /reload and it worked, but now when I reboot it doesn't work. Lemme investigate more.
Edit: Yeah that's weird. /reload and rebooting behave differently. Order of registration issue maybe. It does explain why you thought it worked/didn't work.
Yeah the current version without this PR also works after /reload

sin-ack:
Quick question, then: Would I be able to use `/reload` as a workaround in my current world while this gets fixed?

Mqrius:
Yes should work and not hurt anything!

Mqrius:
Aight the actual issue was trying to iterate over a tag during the kubejs tag registration event. ^That commit instead iterates over a global constant of all fruits that we already have anyway, which fixes the issue.

PR #2675, opened 2026-01-09
Add vinegar as option for curdled milk crafting

JeanRdSz:
This adds an earlier way to make cheese without relying on acetic acid. As vinegar is partially composed of acetic acid, and the recipe for homemade cheese using vinegar is realistic, that would be an useful and accurate recipe.

### DIFF
--- a/kubejs/server_scripts/tfg/food/recipes.food.js
+++ b/kubejs/server_scripts/tfg/food/recipes.food.js
@@ -278,8 +278,8 @@ function registerTFGFoodRecipes(event) {
 
 		event.recipes.gtceu.mixer(`lactose_milk_vinegar_${item.id}`)
 			.circuit(1)
-			.inputFluids(Fluid.of(item.milk, 1000), Fluid.of('tfc:vinegar', 50))
-			.itemOutputs('1x gtceu:lactose_dust')
+			.inputFluids(Fluid.of(item.milk, 1000), Fluid.of('tfc:vinegar', 100))
+			.chancedOutput('gtceu:lactose_dust', 1000, 0)
 			.outputFluids(Fluid.of(item.curdled_fluid, 1000))
 			.duration(300)
 			.EUt(GTValues.VA[GTValues.LV])


Redeix:
I'm debating if it should give lactose dust or not 🤔

Pyritie:
oh you mean in a mixer. Ok, I was worried the TFC cheese-making recipes weren't working

kubejs/server_scripts/tfg/food/recipes.food.js
```diff
@@ -276,6 +276,14 @@ function registerTFGFoodRecipes(event) {
 			.duration(300)
 			.EUt(GTValues.VA[GTValues.LV])
 
+		event.recipes.gtceu.mixer(`lactose_milk_vinegar_${item.id}`)
+			.circuit(1)
+			.inputFluids(Fluid.of(item.milk, 1000), Fluid.of('tfc:vinegar', 50))
```
Pyritie: the ratio of vinegar to acetic acid should roughly match how much acetic acid you can get by distilling vinegar, imo. Otherwise the acetic acid one isn't very useful.

JeanRdSz:
> I'm debating if it should give lactose dust or not 🤔

I mean, lactose dust is useless because you can only craft tablets on MV, as you need Sodium Bicarbonate. But if that an issue, maybe we can make it output just 2 small pile of lactose

Redeix:
> > I'm debating if it should give lactose dust or not 🤔
> 
> I mean, lactose dust is useless because you can only craft tablets on MV, as you need Sodium Bicarbonate. But if that an issue, maybe we can make it output just 2 small pile of lactose

Lactose will be required a lot more in IV so I think it would be better to remove lactose here to keep the acetic acid recipe as more of the advanced option.

kubejs/server_scripts/tfg/food/recipes.food.js
JeanRdSz: I tried to kinda go for the realsitic ratio, but yeah. What if we make it a recipe with 5B of milk to 2B of vinegar? I think that ratio would be fine, and/or remove the lactose dust as Redeix said and put something like milk curds as byproduct instead

Pyritie: what's the ratio that tfc uses? Imo we should match that

JeanRdSz: > what's the ratio that tfc uses? Imo we should match that

TFC uses 9B of milk to 1B of Vinegar

JeanRdSz:
> > I'm debating if it should give lactose dust or not 🤔
> 
> I mean, lactose dust is useless because you can only craft tablets on MV, as you need Sodium Bicarbonate. But if that an issue, maybe we can make it output just 2 small pile of lactose

Ur right. Maybe milk curd would be a better byproduct without losing the initial idea of making cheese easier to make

kubejs/server_scripts/tfg/food/recipes.food.js
Pyritie: I see, so I guess 100mb of vinegar here would be "close enough" because 90mb would feel weird.

PR #2681, opened 2026-01-09
Remove duplicate quest entry from ore processing chapter

KieranBeals:
## Outcome
Fixes: duplicate quest entry

## Additional Information
Old quest page:
image

New quest page:
image

Since there are [no references to this quest](https://github.com/search?q=repo%3ATerraFirmaGreg-Team%2FModpack-Modern%20138C7D4A61E2A77D&type=code), there should be no compatibility issues. 

InboundHalo

### DIFF
--- a/config/ftbquests/quests/chapters/ore_processing.snbt
+++ b/config/ftbquests/quests/chapters/ore_processing.snbt
@@ -176,12 +176,6 @@
 			y: -13.0d
 		}
 		{
-			id: "138C7D4A61E2A77D"
-			linked_quest: "6B10099F3F0931B9"
-			x: 9.0d
-			y: -15.0d
-		}
-		{
 			id: "21670DA706D747EC"
 			linked_quest: "378F0AFCF95354B2"
 			x: 12.0d


Pyritie:
I'm not on my pc rn but can you wash purified dust? I think I remember the quest was duplicated to show all the options you're able to do with the dust at that point

Redeix:
> I'm not on my pc rn but can you wash purified dust? I think I remember the quest was duplicated to show all the options you're able to do with the dust at that point

yeah you can

Pyritie:
then that's why

KieranBeals:
Having a list of parts from 1 through 3 then repeating 3 feels wrong. 
image
image


If we want to keep the last one can we rename it to Early Ore Processing - Part 4?

This would mean making a new title: "{quests.stone_age.clean_dust.title}" and translating it in other languages.

Pyritie:
It's not a step 4 though, it's just a different path you could take. Maybe it would make more sense if the purified ore also had a hammering quest underneath it? I do agree it's inconsistent right now.
In the stone age you can either do crush -> crush -> wash, or crush -> wash -> crush, I think?

KieranBeals:
@Pyritie what made you change your mind?

Pyritie:
I slept on it

PR #2683, opened 2026-01-09
Add Bluemap compat

woopSmile:
Bluemap is a server side webmap, i always use it for my servers. I've made a resources pack to make it compatible with TFG, i dont think it would hurt to include it in the pack. worst case ill open a forum post for the 10 people interested in installing bluemap lol

### DIFF
--- a/.github/ISSUE_TEMPLATE/BUG_REPORT.yml
+++ b/.github/ISSUE_TEMPLATE/BUG_REPORT.yml
@@ -1,5 +1,5 @@
 name: "[🐛]: Bug report"
-description: "Report a bug where something is not working as expected, which does not crash the game."
+description: "Report a bug where something is not working as expected, but does not crash the game."
 title: "[🐛]: "
 labels: [ "Status: Backlog" ]
 projects: ["TerraFirmaGreg-Team/10"]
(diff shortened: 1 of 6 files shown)


Redeix:
upload the resource pack to curseforge and we will pull it from there. Or add the assets to the kubejs folders

woopSmile:
I think this is none of the options? this is not a playable ressources pack for Minecraft, this is only adding tags to blocks (I think?). and for bluemap to consider this resource pack it has to be in config/bluemap/packs which is what I did here. Imma look in bluemap configs if I can change the pack location but I'm not optimistic

Pyritie:
kubejs is basically a big data/resource pack kind of thing. Maybe it could go in /kubejs/config/? I'm not familiar with it. If there's no other options then the zip is fine I guess

woopSmile:
there is no easy way in the config to change the pack folder location. plus, if a user wants to add other resource packs to bluemap, changing the location to kubejs is just gonna cause confusion and fustration, which defeats the purpose of including configs to make it easier on the end user. (in my humble opinion)

Pyritie:
alrighty

Pyritie (approved):

Redeix:
I see. Well after testing it seems to work fine when unzipped so I would prefer if it was that way

Pyritie (changes requested):
change to loose files

PR #2684, opened 2026-01-09
Incorrect tiering for extractor for Bis and Black Bronze

Inceitious:
Bismuth Bronze and Black Bronze wrong tiering in extractor Added some documentation

## What is the new behavior?
Bismuth and Black Bronze have incorrect tiering in extractor. I suspect that this is due to gregtech materials being immutable, so removing blast property for the bronzes only affects the ingot itself and not the rest of the recipe.

## Implementation Details
Added some rules to use LV tier extractor for the bronzes.

## Outcome
Fixed extracting tiering for the bronzes

## Additional Information
I have another idea on how to fix this issue, which is to remove and add in the metal again. But I'm not sure if this will work.

NIL

Inceit


Pyritie:
I feel like it would've been cleaner to pull out the check into a variable at the top of the method, instead of copying and pasting the same block everywhere. I also think it would be good to bring this issue up to GT because yeah it's confusing me too.

Inceitious:
> I feel like it would've been cleaner to pull out the check into a variable at the top of the method, instead of copying and pasting the same block everywhere. I also think it would be good to bring this issue up to GT because yeah it's confusing me too.

Sure, I can do the check into variable no problem.

Inceitious:
Edited. If anyone wants to add in more Bismuth Bronze or Black Bronze items for extractor, they can use BisBlackCheck(material). For awareness.

Pyritie:
ah sorry, I should've been more clear, I meant like this at the top:

```js
const extractorEUt = material.hasProperty(PropertyKey.BLAST) && material !== GTMaterials.BismuthBronze && material !== GTMaterials.BlackBronze
	? GTValues.VA[GTValues.MV]
	: GTValues.VA[GTValues.ULV]
```

and then you can just have the single extractor calls as before, like this:
```js
event.recipes.gtceu.extractor(`gtceu:extract_${material.getName()}_${tagPrefixName}`)
	.itemInputs(toolHeadItem)
	.outputFluids(Fluid.of(material.getFluid(), materialAmount * 144))
	.duration(material.getMass() * 6 * materialAmount)
	.category(GTRecipeCategories.EXTRACTOR_RECYCLING)
	.EUt(extractorEUt)
```
I think that would be cleaner, what do you think?

Inceitious:
> ```js
> const extractorEUt = material.hasProperty(PropertyKey.BLAST) && material !== GTMaterials.BismuthBronze && material !== GTMaterials.BlackBronze
> 	? GTValues.VA[GTValues.MV]
> 	: GTValues.VA[GTValues.ULV]
> ```

Looks good, but I do have a question on the tiering. If you put ULV there, does this mean that steam machines are able to use these recipes? Else, I can put it at LV tier instead.

Pyritie:
It's confusing, but steam machines can do LV recipes as long as they don't involve fluids. Since all these extractor recipes involve metal fluids, the steam one can't do it, so it's not a problem

Pyritie:
almost! I think `extractorEUt` can go right near the top of the method, so it's only made once? instead of in multiple places in the file? sorry for being so picky lol

Inceitious:
> almost! I think `extractorEUt` can go right near the top of the method, so it's only made once? instead of in multiple places in the file? sorry for being so picky lol

I'm not sure whether that will work since each block has it's own parameters for the materials, so I don't think I can put it once at the top unless I do a function

Inceitious:
It's okay to be picky haha! I like being organised too. Speaking of being organised, I took the liberty to change the tiering for the fluid solidifier, arc furnace, and assembler to LV tier, since some of the other recipes uses LV tiering system to them (e.g. Bismuth Bronze Frame) and I like consistency too. I hope you don't mind that! Didn't touch the macerator or alloy smelter in case anything goes wrong.

Pyritie:
> > almost! I think `extractorEUt` can go right near the top of the method, so it's only made once? instead of in multiple places in the file? sorry for being so picky lol
> 
> I'm not sure whether that will work since each block has it's own parameters for the materials, so I don't think I can put it once at the top unless I do a function

ooh right. Function sounds good then I think? Like a `getExtractorEUt(material)` that the recipe can just call directly

> Speaking of being organised, I took the liberty to change the tiering for the fluid solidifier, arc furnace, and assembler to LV tier, since some of the other recipes uses LV tiering system to them (e.g. Bismuth Bronze Frame) and I like consistency too.

That should be fine!

Inceitious:
Somewhat cleaned up the tiering system. Same as before, didn't touch any alloy smelters, macerators, or any steam machines. Let me know if there are any issues.

Pyritie:
ooh you meant like that... I thought you just meant in that file. Changing the whole everything is too much. While a few of them probably could be changed, the thought process was "anything that could be done in a crafting grid (just with a tool) should be ULV" or "anything that you can do with a tfc thing like a vat or whatever should be ULV" if that makes sense? 🤔Especially for the cutter, because that also affects the Greate mechanical saw recipe tiers. And the macerator should always be ULV for everything.

Inceitious:
> ooh you meant like that... I thought you just meant in that file. Changing the whole everything is too much. While a few of them probably could be changed, the thought process was "anything that could be done in a crafting grid (just with a tool) should be ULV" or "anything that you can do with a tfc thing like a vat or whatever should be ULV" if that makes sense? 🤔

Hm, I was thinking that machines that have only LV tiering should be limited to LV tier as it is 1) easier to understand for newcomers as they won't think that there is a ULV version for that machine 2) sign of a good modpack where everything is detailed 3) preference

> Especially for the cutter, because that also affects the Greate mechanical saw recipe tiers. And the macerator should always be ULV for everything.

You're right. In that case I will just revert this commit and submit a new PR next time, after I know which ones to touch on ( I'm guessing I can't touch Mixer, Cutter, Washer?, Extruder and all the relevant greate machines too)

Pyritie:
> Hm, I was thinking that machines that have only LV tiering should be limited to LV tier as it is 1) easier to understand for newcomers as they won't think that there is a ULV version for that machine

the confusion I see the most often is people thinking their steam machine can't do an LV recipe. I'm not really sure what to do about that though, other than putting it in the machine tooltip somewhere or something. It's a weird gregtech quirk that confuses people in every pack, not just ours.

Pyritie:
By the way, I don't know if this will change anything at all, but I did see that these bronzes still had an ALLOY_BLAST property as well. Removing the property didn't remove their "molten" however, so I'm not sure what's going on.

Inceitious:
> By the way, I don't know if this will change anything at all, but I did see that these bronzes still had an ALLOY_BLAST property as well. Removing the property didn't remove their "molten" however, so I'm not sure what's going on.

I suspect it's the same issue as well. When we remove the properties of an item A, it will only affect A itself and not Aa (item generated from A). Probably because GT material registration comes first, then kubejs next.

Pyritie:
I've been removing properties from other materials as part of my work on stone dusts. Removing the GEM property from glass, for example, does remove all the different glass gems. However, removing the INGOT property from thorium will remove nuggets, ingots, and blocks, but _not_ plates and rods. It doesn't look like there's a way to remove _flags_, which is what generates the plates and rods. Maybe the molten fluid is the same.

Pyritie:
In case this is useful, there's also a few items that you can't remove properties from without GT complaining. For example, you can't remove FLUID from Blaze, or remove DUST from Flour or any of the stone types. Removing ARMOR also doesn't work.

Redeix (approved):

Issue #2647, opened 2026-01-07
[🪡]: Add more solders


Pyritie:
### Was this enhancement asked before?

- [x] No, it was not.

### What would you like to be added?

Remove tin as a solder option earlier, maybe around MV or so.

Add a new higher tier solder for higher tier crafts.

Optionally, add another solder that uses bismuth as another use for it (I think it's called wood's alloy or something)

### Why is this needed?

More resource sinks more gooder

SpicyNoodle5:
Wood's metal seems to be roughly 50% Bi, 27% Pb, 13% Sn, 10% Cd (yay another cadmium use)
There's also similar alloys with added indium or indium + thallium (new element!?)

Not sure when you'd want to introduce new alloys, perhaps it could be interesting to get a new solder every couple circuit tiers? Maybe something like:
Early circuits: tin/soldering alloys
Micro/Nano: soldering alloy/woods metal
Quantum/Crystal: woods metal/higher tier alloy (maybe the alloy with added indium + thallium?)
Wetware: higher tier alloy only

Just throwing ideas out there

PR #2690, opened 2026-01-09
Add New Solders

SpicyNoodle5:
## What is the new behavior?
#2647 - seemed interesting to look into, open to any other suggestions on this.

## Implementation Details
There might be a better way to handle these replaceInputs.

## Outcome
Adds 2 new solders: Wood's Metal (Bi Pb Sn Cd) and Bi-Pb-Sn-Cd-In-Tl alloy.
Tiers soldering alloys so every other circuit tier from micro processors introduces a new solder and each circuit tier has a "cheaper" alternative:
Early circuits: Tin/Soldering alloy
Micro/Nano: Soldering alloy/Wood's metal
Quantum/Crystal: Wood's metal/Bi-Pb-Sn_Cd_In_Tl
Wetware: Bi-Pb-Cd-In-Tl

Adds a new element - Thallium, which will need an ore. Potential candidates could be crookesite (TlCu5Se4) or hutchinsonite (TlPbAs5S9) or apparently in trace amounts from pyrite.


Pyritie:
is that last one even a real alloy? lol

kubejs/server_scripts/tfg/machines/recipes.circuits.js
Pyritie, line 1: it'd be good to move all the other circuit changes into this one file, to keep everything together

Pyritie:
also, given that it's an option right around IV, thallium sounds like another good candidate for something exclusively from venus oreproc (like zirconium will be)

SpicyNoodle5:
> is that last one even a real alloy? lol

Apparently, its listed here as related to Wood's metal: https://en.wikipedia.org/wiki/Wood%27s_metal
Supposedly it has a very low melting point at around 41C

Pyritie:
@TomPlop I'd like your input here too -- we can either wait on this PR until venus, add it now and put thallium somewhere temporary, or just comment out that tier of solder until later

kubejs/server_scripts/tfg/machines/recipes.circuits.js
SpicyNoodle5: done

kubejs/startup_scripts/tfg/materials.js
```diff
@@ -198,4 +198,31 @@ const registerTFGMaterials = (event) => {
...
+	event.create('tfg:woods_metal')
+		.dust()
+		.liquid(GTFluidBuilder().temperature(343))
+		.ingot()
+		.color(0x65737C)
+		.components('4x bismuth', '2x lead', '1x tin', '1x cadmium')
+		.flags(GTMaterialFlags.DECOMPOSITION_BY_CENTRIFUGING)
+		.hazard(HazardProperty.HazardTrigger.SKIN_CONTACT, GTMedicalConditions.POISON)
+
+	event.create('tfg:thallium')
```
Pyritie: oh also, gregtech actually comes with materials for every element built-in. Look at what we do to make zirconium "visible" (I think we add the dust property to it?) instead of making a new material

SpicyNoodle5: ah interesting, was wondering how GT seemed to already know that thallium existed lol

SpicyNoodle5:
> for the zpm+ recipes, wouldn't this still have two recipes, both with the thallium solder but one with fewer mB than the other?

In game it seems to just keep one of the recipes but yeah I'll change this to make it clearer

Pyritie (dismissed):

TomPlop:
I think it's really cool but I would prefer if we keep this one on hold until we begin the rework on MV and HV. My main issue is that the new soldering alloys are a bit basic to make and don't really make use of new mechanics.

The implementation is great though so just keeping it on the side until we start a bit more delving into the rework and also find more interesting ways to do the alloys than just a mixer.

Xtrial-01:
Bi-Pb-Sn-Cd-In-Tl alloy might actually not be suitable for wetware due to its toxicity, gt wetware needs special sterilized and cleaned environment so it seems to be quite fragile

SpicyNoodle5:
Didn't even think about that lol, tho tbf regular soldering alloy has lead in it so arguably isn't much better. I guess if we wanted to, it could be swapped out for something else later (maybe a fluid from the flesh planet whenever ZPM gets worked on or something)? It could still be used for other recipes in that tier which reminds me that the assembly line recipes for things like motors still use regular soldering alloy so I could add the tiering to that as well.

SpicyNoodle5:
Made some changes to this, apologies for leaving it for a while.

Moved all solder tiers down by 1 to start at MV, kept wood's metal for the MV/HV solder and the thallium alloy for EV/IV. Currently there isn't a new solder for LuV+.

Process for wood's metal: Mixer -> Dust -> Pyro -> Fluid

Process for thallium: OLA -> Lorandite -> LCR + Sulfuric Acid -> Thallium Sulfate -> Chem Reactor + Zinc -> Thallium. The zinc sulfate produced can be electrolyzed to close the loop. The alloy is made in the ABS. This also uses indium which should be fine since it can be made at EV (even if its usually an IV material).

There's probably balancing things to address, mainly the OLA recipe. Lorandite also acts as a source of arsenic which could be a potential issue (all thallium ores seem to contain some arsenic). Could also make Lorandite a dusty ore to keep consistency in the OLA (just wasn't sure on how to do that).

Pyritie:
> Could also make Lorandite a dusty ore to keep consistency in the OLA (just wasn't sure on how to do that).

add `TFGMaterialFlags.GENERATE_DUSTY_ORES` to the material flags

TomPlop:
@SpicyNoodle5 Can you fix conflict for this PR for MV/HV rework branch

kubejs/server_scripts/tfg/machines/recipes.circuits.js
```diff
@@ -0,0 +1,342 @@
...
+		.itemInputs('8x #forge:dusts/woods_metal')
+		.inputFluids(Fluid.of('gtceu:nitrogen', 1000))
+		.outputFluids(Fluid.of('tfg:woods_metal', 1152))
+		.duration(20 * 48)
+		.circuit(2)
+		.EUt(GTValues.VA[GTValues.MV])
+    
+	// BiPbSnCdInTl
+	
+	event.recipes.gtceu.ostrum_linear_accelerator('tfg:lorandite_ola')
```
TomPlop: Would prefer to not have that one for now. You can just comment it out

SpicyNoodle5: Done

Pyritie:
#3166

Issue #2691, opened 2026-01-09
[🪶]: heated mixer recipe for latex/conifer pitch + wood ash into sticky resin


BadsyEdgecat:
### Was this feature suggested before?

- [x] No, it was not.

### Version modpack used

0.11.16

### Your solution

would be nice if a heated mechanical mixer could be used to make sticky resin (and possibly vulcanized latex) instead of exclusively in the vat until gt machines.

JeanRdSz:
Honestly, it's a good concept. But I think this recipe would fit better in the Vacuum Chamber

PR #2693, opened 2026-01-09
Mixer recipes for recipes with latex

JeanRdSz:
This will add a way to make recipes that utilize latex on the Vacuumizer, following the sugestion made on #2691.

To be honest, this will make the vat a bit less useful after you make progress on Greate. But It's a step up on automation and would be nice to have options before LV machines.


Pyritie:
tbh, I'd prefer the mixer if only because it's a bit later in steam age

Pyritie (approved):

PR #2705, opened 2026-01-11
Added alloy smelter recipes for jar lids

thederpysockdude123:
adds alloy smelter recipes for jar lids, as well as rebalancing the jar lid recipes for consistency.

decreased anvil yield of standard TFC jar lids from 16 to 8, and increased the yield of the fluid solidifier recipe to 24 to align with stainless steel jar lids giving 16 on the anvil and 48 in the fluid solidifier (thus 3x as much). changes reflected in alloy smelter recipes.

Federation President Bravo


kubejs/server_scripts/tfc/recipes.metals.js
```diff
@@ -303,6 +303,11 @@ function registerTFCMetalsRecipes(event) {
 		.tier(3)
 		.id(`tfc:anvil/iron_door`)
 
+		// jar lid
+	event.recipes.tfc.anvil('8x tfc:jar_lid', '#forge:ingots/tin', ['hit_last', 'hit_second_last', 'punch_third_last'])
```
Pyritie: given that people are mostly going to be interacting with jars pre-steam, I don't think the anvil recipe needs to be nerfed. Tin is already a pain to work with on the anvil lol

thederpysockdude123: fair enough, went ahead and changed it to 16 then, probably could just delete it and leave it as the default TFC recipe as well

kubejs/server_scripts/tfc/recipes.metals.js
```diff
@@ -304,7 +304,7 @@ function registerTFCMetalsRecipes(event) {
 		.id(`tfc:anvil/iron_door`)
 
 		// jar lid
-	event.recipes.tfc.anvil('8x tfc:jar_lid', '#forge:ingots/tin', ['hit_last', 'hit_second_last', 'punch_third_last'])
+	event.recipes.tfc.anvil('16x tfc:jar_lid', '#forge:ingots/tin', ['hit_last', 'hit_second_last', 'punch_third_last'])
```
Pyritie: isn't this just the same as what it was originally now? you can just delete this recipe and the `event.remove()` for it in the other place then lol

thederpysockdude123: done 👍

Redeix (approved):

PR #2733, opened 2026-01-13
Feature/tidying recipes

Inceitious:
## What is the new behavior?
Recipe bloat for signs recycling in EMI.

## Implementation Details
Previously, there was a section that says to fix sign metal dupes for recycling recipe, and there was ```event.remove(`gtceu:arc_furnace/arc_wood/hanging_sign/${metal.getName()}/${wood}`)``` implemented. I think how I wrote it removes the need for this line, probably.

## Outcome
Added tags to support hanging signs recycling. Edited the hanging signs recycling recipes.

## Additional Information
NIL

Your new refactor Pyritie, not sure if its compatible.

Inceit


kubejs/server_scripts/tfc/tags.js
```diff
@@ -434,6 +434,35 @@ function registerTFCItemTags(event) {
...
+        GTMaterials.BismuthBronze,
+        GTMaterials.WroughtIron,
+        GTMaterials.Steel,
+        GTMaterials.BlackSteel,
+        GTMaterials.RedSteel,
+        GTMaterials.BlueSteel
+    ];
+
+    SIGN_METALS.forEach(metal => {
+        event.add(`tfg:hanging_sign/${metal.getName()}/hardwood`, global.TFC_HARDWOOD_TYPES.map(wood => `tfc:wood/hanging_sign/${metal.getName()}/${wood}`))
```
Pyritie: these four loops can all be merged into one loop, right?

kubejs/server_scripts/tfc/tags.js
```diff
@@ -434,6 +434,35 @@ function registerTFCItemTags(event) {
     event.add('forge:dyes/green', 'tfc:powder/malachite')
     event.add('forge:dyes/brown', 'tfc:powder/garnierite')
     event.add('forge:dyes/yellow', 'tfc:powder/limonite')
+
+    //Hanging Signs Soft & Hard Wood
+    const SIGN_METALS = [
```
Pyritie: you could use `global.TFC_EQUIPMENT_METALS`, it has the same list

kubejs/server_scripts/gregtech/recipes.recycling.js
```diff
@@ -128,81 +128,55 @@ function registerGTCEURecyclingRecipes(event) {
...
-				.EUt(8)
-				.category(GTRecipeCategories.MACERATOR_RECYCLING)
-		})
-
-		global.AFC_HARDWOOD_TYPES.forEach(wood => {
-			event.recipes.gtceu.macerator(`gtceu:macerator/macerate_wood/hanging_sign/${metal.getName()}/${wood}`)
-				.itemInputs(`afc:wood/hanging_sign/${metal.getName()}/${wood}`)
+		
+		// Macerating
+		global.AFC_HARDWOOD_TYPES.concat(global.TFC_HARDWOOD_TYPES).forEach(wood => {
```
Pyritie: don't you remove the `forEach` since you use tags now?

Pyritie (changes requested):

kubejs/server_scripts/gregtech/recipes.recycling.js
Inceitious: What do you mean?

Pyritie: well, you're still looping over all of the wood types

Inceitious: I used `TFC_HARDWOOD_TYPES` and `TFC_SOFTWOOD_TYPES`, if that is what you mean. Otherwise, I can move the macerator recipes into the loop for each one

Pyritie: do you need the loop at all? wouldn't it just be 1 loop for all the metals, then inside that, 2 macerator and 2 arc furnace recipes (1 of each for each wood type), since you have the tags?

like this:
```
for each metal:
    macerate softwood
    arc furnace softwood
    macerate hardwood
    arc furnace softwood
```

Pyritie (dismissed):

Pyritie:
does this make more sense?

Inceitious:
> does this make more sense?

Yeah, it does.

Actually, that's what I did originally for the wood array portion lol. I only added the array because it didn't work the first time round for some reason, and after I changed some codes which includes adding that, it worked.

Probably mistook that code to be the success factor when something else is probably the solution.

Pyritie:
oh, all my changes were overwritten now 😂

Inceitious:
> oh, all my changes were overwritten now 😂

Uh oh, lemme fix that

Issue #2682, opened 2026-01-09
[❓]: Could More Red Mod Content Be Integrated into the "ComputerCraft" FTB Quest Line?


hawwwa:
### Was this question asked before?

- [x] No, it was not.

### Your Question

The More Red mod's items can be crafted at the LV stage, but their practical uses are not intuitive—I'm currently unsure how they are meant to be used, especially in combination with other mods. Additionally, some players may not even be aware this mod is included. Adding a dedicated series of quests within the existing “ComputerCraft” chapter to introduce More Red would therefore be very helpful.

CapitalismWouldveWorked:
I have alternative suggestion: After the electron tubes quest, add a tab dedicated to redstone devices and use cases (More Red stuff, Create redstone, GT covers, etc) with computercraft content being a part of it.

Pyritie:
would you be interested in helping to write them?

hawwwa:
I can write these, but I might need some time first to figure out how to add this kind of content to Patchouli.I've noticed that the patchouli_books folder within the kubejs directory appears to hold just the translation files.

Pyritie:
The patchouli books are for the field guide, the quests are in /config/ftbquests. Quests have an in-game editor though. Go into creative (or turn on cheats) and there should be a little pencil button on the bottom right. Look at how other quests put their text in, with language IDs instead of english text. This is to support translations, so you'll also have to send your text over [here](https://github.com/TerraFirmaGreg-Team/Tools-Modern/tree/dev/LanguageMerger/LanguageFiles/tfg/en_us/Quests).

I wouldn't mind if you just made a separate chapter either. There's already someone else working on redoing the CC chapter from scratch.

PR #2739, opened 2026-01-14
adjustments to CC quests

hawwwa:
Adjusted the quest layout in the CC chapter and added the corresponding description placeholders. As per #2682, the actual English text will be contributed separately to the Tools-Modern repository.

### DIFF
--- a/config/ftbquests/quests/chapters/cc.snbt
+++ b/config/ftbquests/quests/chapters/cc.snbt
@@ -48,7 +48,7 @@
 					type: "item"
 				}
 			]
-			title: "{quests.comouter.disks.title}"
+			title: "{quests.computer.disks.title}"
 			x: 0.0d
 			y: 1.5d
 		}
@@ -137,7 +137,7 @@
 				"502B24C71210D52E"
 			]
 			dependency_requirement: "one_completed"
-			description: ["{quests.computer.turtle_advancdd.description}"]
+			description: ["{quests.computer.turtle_advanced.description}"]
 			id: "61B4EDA4365E1B7A"
 			tasks: [{
 				id: "4032A8D2000CDBAA"


Pyritie:
do you have a picture of what the new layout looks like? I can't really visualise the text file lol

hawwwa:
> 你有新布局的照片吗？我真的无法想象那个文本文件，哈哈

I placed the Wireless Modem in the middle part of the quest flow mainly because it can connect various computer devices. Since there aren't many items involved, the quest line couldn't be extended too long like some others. As for the More Red mod—I can't exactly cram a bunch of digital circuit theory into a quest description, so I left it out.
屏幕截图 2026-01-14 183231
2026-01-14 18_59_04-TerraFirmaGreg-Modern

Jeuvke:
Some critiques:
- The mod is called CC: Tweaked (CC:T), to differentiate itself from Dan's original mod.
- Where are the quests for all the other peripherals? The peripherals from Advanced Peripherals aren't even mentioned.
- Floppies are mostly useless, as you can just put an entire computer in a disk drive instead. You also didn't mention downloading programs from github or pastebin, nor did you mention the drag-and-drop feature to copy programs to a computer from your own physical computer.
- The turtle quest mentions you can equip tools to use them. That is not the case in TFG, as there are no vanilla diamond tools here. You _can_ use tools with turtles by using AP's automata core, but that's not mentioned anywhere.
- The wireless modem quest never mentions GPS, which is the number 1 usecase for wireless modems.
- The quests mentioning redstone do not mention that CC:T has integration with more red's bundled cables.
- There are no code examples at all.
- There are no links to any documentation, for either CC:T, AP, or just lua itself.

hawwwa:
Apologies, my initial idea was just to write a simple task guide. I’ll need some time to reorganize this content, but I’ll make it better than before.

Pyritie:
Jeuvke's been working on their own quest chapter for CC, but I believe it
doesn't go into morered at all, so if you'd like to focus on morered
instead, that will reduce the overlap.

If you have any ideas for places in TFG that morered is really useful for,
that would be helpful too, because right now we don't really use redstone
for much.

PR #2808, opened 2026-01-22
added Ad Astra airlock to valid cleanroom doors

thederpysockdude123:
added the Airlock from Ad Astra to the list of valid doors for the Cleanroom as it was missing.

Federation President Bravo


Mqrius:
Isn't the ad astra airlock door 3x3? So it doesn't work in a cleanroom cos you can have max 8 door blocks? I'm not opposed to making that 9 blocks, airlock door would be nice.

Pyritie:
> Isn't the ad astra airlock door 3x3? So it doesn't work in a cleanroom cos you can have max 8 door blocks? I'm not opposed to making that 9 blocks, airlock door would be nice.

we already have a mixin to allow 9 blocks!

PR #2922, opened 2026-01-31
Add bluemap support v2

woopSmile:
This is the bluemap pr but unzipped, didnt know it worked without being zipped, this is just better. 
Gregtech blocks still need love to work with bluemap, but at least the entire world wont be gray


Redeix:
Yan is getting conflicting results. So it might not be possible for everyone. But we can merge this and zip it ourselves and resubmit. Having the files viewable in commit history is good enough for me.

Redeix:
I apologize for the trouble. We just prefer to keep everything viewable on GitHub if we can

woopSmile:
no trouble caused. having a zip file is not ideal and i was scared it was gonna cause issues with publishing to curseforge, the only reason it was a zip is that i thought it was necessary. as for yan, i place a bet that he did a mistake on his end (hopefully i didn't jinxed myself)

Redeix:
yeah curseforge is my concern too since I think they complained about other miscellaneous files we had before too. Are you able to verify if having the files unzipped worked for you too?

woopSmile:
tested before pr, file unzipped works on my machine.

PR #2946, opened 2026-02-02
Fix dimensionalPowerConsumption method overload

Mqrius:
There was never actually any overload? dimensionalPowerConsumption aka TFGAE2PowerConsumption only has one `add` method, and TFGAE2PowerConsumption extends EventJS which doesn't have any `add` method. It just works like this.

I added a value for Venus since it's coming up soon.

Goes with https://github.com/TerraFirmaGreg-Team/Core-Modern/pull/273 though they're both standalone.

### DIFF
--- a/kubejs/server_scripts/tfg/events.js
+++ b/kubejs/server_scripts/tfg/events.js
@@ -1,14 +1,15 @@
 "use strict";
 
-// TODO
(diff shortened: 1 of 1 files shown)


Pyritie:
I guess kolja already removed the overload and we forgot to update this file

PR #3002, opened 2026-02-07
Refactor/wood

Inceitious:
## What is the new behavior?
Refactoring of wood recipes.

## Implementation Details
First iteration. 

## Outcome
All wood recipes can be found and be edited easier in the future.

## Additional Information
NA.

Hopefully none.

Inceit

### DIFF
--- a/kubejs/startup_scripts/tfg/worldgen/items.wood.js
+++ b/kubejs/startup_scripts/tfg/worldgen/items.wood.js
@@ -26,5 +26,4 @@ function registerTFGWoodItems(event) {
 	event.create('tfg:wood/lumber/bamboo')
 	.tag('tfc:lumber')
 	.texture('tfg:item/wood/lumber/bamboo')
-	.displayName('Bamboo Lumber')
 }


Xtrial-01:
i can't approve review but just want to say that this is so cool

kubejs/server_scripts/tfg/recipes.js
```diff
@@ -44,12 +44,14 @@ const registerTFGRecipes = (event) => {
...
 	registerTFGRubberRecipes(event)
 	registerTFGWoodRecipes(event)
+	registerTFGWoodenRecipes(event)
 	registerTFGKnappingRecipes(event)
 	registerTFGLeatherRecipes(event)
 	registerTFGClayRecipes(event)
 	registerTFGEquipmentRecipes(event)
 	registerTFGDyeRecipes(event)
 	registerTFGArtisanTableRecipes(event)
+	registerTFCScrapingKnivesRecipes(event)
```
Pyritie: why was this moved?

Inceitious: I felt it will be more suitable to consolidate all the recipes that tfg changes, which included tfcscrapingknives, into tfg server script, instead of the main server script.

Pyritie: no, because that would include almost all of the recipes we have

Inceitious:
> This isn't quite what I had in mind -- I was more thinking of like how the rocks are done, where there's one big table that has ALL of the various woods in it, and then one file that loops through the table and generates the recipes so they'll always be the same.

There is a woodbuilder function, which generates the recipes from the different mods and keeps it consistent. 
> 
> Your change just looks like it moved all the woods from different mods into the same file. Which ok, helps a little, but it still doesn't ensure consistency like a big table would.

kubejs/server_scripts/tfg/recipes.js
Inceitious: Got it, I'll move it back next iteration. So TFGrecipe server script is just for the recipes within the tfg.js file correct?

Pyritie: yeah

Pyritie:
I've read it a second time.
I think I misunderstood some of your changes.
Maybe it would be a bit clearer if all of the `woodBuilder` methods were together? And then all of the removes and miscellaneous recipes were also in their own parts of the file.

Inceitious:
> I've read it a second time. I think I misunderstood some of your changes. Maybe it would be a bit clearer if all of the `woodBuilder` methods were together? And then all of the removes and miscellaneous recipes were also in their own parts of the file.

By in their own parts of the file, do you mean the original place they belonged to? For example the removed axle from beneath, you meant to move them back to the beneath.js file and keep just the woodbuilder in tfg.recipes.wood.js?

Pyritie:
No, I think it's good to keep everything in your recipes.wood.js. I mean like, put the woodbuilder uses at the top, then all the miscellaneous recipes, then all of the removes. Or something like that. What do you think?

Inceitious:
> No, I think it's good to keep everything in your recipes.wood.js. I mean like, put the woodbuilder uses at the top, then all the miscellaneous recipes, then all of the removes. Or something like that. What do you think?

Oh, I was considering that way too, but I felt that separating by mod types would be clearer, even if it is not as clean. I feel that it is easier to find a recipe that is regionised by mods (since I demarcated where each mod is) rather than finding a recipe that is regionised by type of recipe. 

And also I don't have to write global.wood_types for each mod removal lol

Pyritie:
true, let's continue then. We can always rearrange things later if we change our mind

Inceitious:
Figured out the issue with the outliers, time for a 2nd round of PR

kubejs/startup_scripts/tfg/worldgen/items.wood.js
```diff
@@ -21,4 +21,10 @@ function registerTFGWoodItems(event) {
 	event.create('tfg:twigs/alphacene')
 		.tag('tfc:twigs')
 		.texture('tfg:item/wood/twig/alphacene')
+
+	// Bamboo lumber
+	event.create('tfg:wood/lumber/bamboo')
+	.tag('tfc:lumber')
+	.texture('tfg:item/wood/lumber/bamboo')
+	.displayName('Bamboo Lumber')
```
Pyritie: this will need a lang string created in Tools-Modern, I can show you how to do that if you haven't done it before

Inceitious: Yes please, how do I do that?

Pyritie: You'll need to PR into [this file](https://github.com/TerraFirmaGreg-Team/Tools-Modern/blob/dev/LanguageMerger/LanguageFiles/tfg/en_us/items.json) to add your item, so other translators can translate it as well. You can delete the `.displayName()` here. In this case, the lang ID would be `item.tfg.wood.lumber.bamboo`

Redeix (approved):

PR #3016, opened 2026-02-09
Add Bags for Universal Composts

Nebby1999:
## What is the new behavior?
Added two new items, which act as Bags of Universal Composts.
These bags are not meant to improve the storage experience (they stack up to 16), but rather, improve the early game usage of them, Bags of Universal Composts are equal to the "High Compost" items, which means that to fill a compost you only need to click eight times instead of 32 times.

Two new quests have been added to explain the composter and universal composts.

Bags of composts can be crafted into four universal composts if desired

## Implementation Details
Implemented with KubeJS, not much to say.
Art is kinda meh, but it works

## Outcome
The user experience of using the Universal Composts in early game should improve, granted this doesnt change anything once you get the mixer, but again, it's meant for early game

## Additional Information
image
image

None known

Haha, once again contributing once in a blue moon 🔥

### DIFF
--- a/kubejs/assets/ambientsounds/basic/regions/jungle.json
+++ b/kubejs/assets/ambientsounds/basic/regions/jungle.json
@@ -12,7 +12,19 @@
         "ocean_reef",
         "deep_ocean",
         "deep_ocean_trench",
-        "rain"
+        "rain",
+        "sangnum_edge",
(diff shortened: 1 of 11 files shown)


Nebby1999:
Worth mentioning, the tools repo has already been updated with the tokens for this feature

PR #3037, opened 2026-02-11
Weird ass bug potential fix

Coox1e:
Nitrogen and Phosphorus werent working (fertilizer values remain at 10%) but potassium was. Only difference is where the variable is, and what it was followed by. Potassium worked potentially because it didnt have a comma?? This might fix it????? this bug should not occur but it does. i am an evil developer that does not test my simple code

### DIFF
--- a/kubejs/server_scripts/tfc/data.js
+++ b/kubejs/server_scripts/tfc/data.js
@@ -110,9 +110,9 @@ const registerTFCFertilizers = (event) => {
 	event.fertilizer('gtceu:small_ammonium_chloride_dust', 0.075, null, null)
 	event.fertilizer('gtceu:ammonium_chloride_dust', 0.3, null, null)
 	
-	event.fertilizer('tfc:pure_nitrogen', 1, null, null)
-	event.fertilizer('tfc:pure_phosphorus', null, 1, null)
-	event.fertilizer('tfc:pure_potassium', null, null, 1)
+	event.fertilizer('tfc:pure_nitrogen', 1.0, null, null)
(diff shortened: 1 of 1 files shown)


gustovafing:
This makes no difference for numbers in JavaScript.

Coox1e:
i know but also the position of variables shouldnt either

Coox1e:
why does 1 only work in one of the three spots

Pyritie:
Since these items already had fertiliser values, it's possible that they've each now got _two_, and tfc just picks one of them.
What you could do instead, is find the data file where these get registered in tfc, and put files in the identical locations inside the kubejs/data/ folder. That will properly overwrite them

Pyritie:
(unless this ends up being a weird bug like with drinkables...)

Coox1e:
i guessed it might be this, its just throwing me off that ONE worked but the other two didnt

Coox1e:
may just be a load order error

Pyritie:
yeah exactly

Coox1e:
pure fertilizers are only mentioned in two files in this repository according to a comprehensive search...

Pyritie:
> pure fertilizers are only mentioned in two files in this repository according to a comprehensive search...

https://github.com/TerraFirmaCraft/TerraFirmaCraft/tree/1.20.x/src/main/resources/data/tfc/tfc/fertilizers

Coox1e:
my search was not comprehensive enough... thank you

Coox1e:
beg please delete those next update okay...

PR #3045, opened 2026-02-12
Datagen

gustovafing:


kubejs/assets/tfg/models/block/mars_icicle.json
Pyritie, line 1: I don't understand why these needed to be moved to datagen when there's only a single other block that uses this model

Pyritie (dismissed):

PR #3011, opened 2026-02-08
Add redstone etching fluid

Xtrial-01:
Adds Redstone Etching Fluid for HV 

For any details check mv/hv rework thread on devchat, everything about it was discussed there

I will pr langs for new materials soon

PR #3049, opened 2026-02-12
Mv hv rework redstone fluid (reopened because git moment)

Xtrial-01:
Adds redstone etching fluid to hv, as specified in dev chat and on original pr (#3011)

Had to reopen because github nuked my first pr for no particular reason (i think that renaming the origin branch might be the fault but idk)

<sup>why its always so stubborn towards me bruh</sup>


Pyritie:
yeah, renaming branches will do that

PR #3055, opened 2026-02-13
fixed typo in LV initial quest description

Eilandor:


Eilandor:
removed extra "will" word:

"... Our quests will try to will guide you ..." -> "... Our quests will try to guide you ..."

Pyritie:
Thanks, would you be able to make the same change [over here](https://github.com/TerraFirmaGreg-Team/Tools-Modern/blob/dev/LanguageMerger/LanguageFiles/tfg/en_us/Quests/low_voltage.json) as well? we use a tool to manage all of our lang files, so your change will be overwritten if it's not over there too

Eilandor:
yes. created PR in that repository too.

Pyritie (approved):
thanks!

PR #3159, opened 2026-02-22
Add optional recipes for Create: Extra Gauges

TreyR9:
I want it to be added to main but here's the next best thing
As per usual yell at me if this breaks something

### DIFF
--- a/kubejs/server_scripts/create_extra_gauges/recipes.js
+++ b/kubejs/server_scripts/create_extra_gauges/recipes.js
@@ -2,160 +2,160 @@
 "use strict";
 
 const registerCreateExtraGaugesRecipes = (event) => {
-if (Platform.isLoaded("extra_gauges")) {
-	event.remove({ mod: 'extra_gauges' });
+	if (Platform.isLoaded("extra_gauges")) {
+		event.remove({ mod: 'extra_gauges' });
 
-	event.shaped(
-		Item.of('extra_gauges:logic_gauge'),
-		[
-			'AEA',
-			'ECE',
-			' F '
-		],
-		{
-			A: 'minecraft:redstone_torch',
-			C: 'create:factory_gauge',
-			E: 'gtceu:fine_red_alloy_wire',
-			F: 'create:electron_tube'
-		}
-	)
+		event.shaped(
+			Item.of('extra_gauges:logic_gauge'),
+			[
+				'AEA',
+				'ECE',
+				' F '
+			],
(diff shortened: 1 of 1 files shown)


kubejs/server_scripts/create_extra_gauges/recipes.js
```diff
@@ -0,0 +1,178 @@
...
+		[
+			'FHF',
+			'ECE',
+			' E '
+		],
+		{
+			C: 'create:factory_gauge',
+			E: 'gtceu:fine_red_alloy_wire',
+			F: 'create:electron_tube',
+			H: 'create:precision_mechanism'
```
Pyritie: gauges already require a precision mechanism, a second one will bump up the cost a lot, is that intentional?

kubejs/server_scripts/create_extra_gauges/recipes.js
```diff
@@ -0,0 +1,178 @@
...
+		}
+	)
+	
+	//event.recipes.gtceu.shaped('extra_gauges:passive_gauge',
```
Pyritie: do these comments still need to be here?

TreyR9: ffs version control is hard

kubejs/server_scripts/create_extra_gauges/recipes.js
TreyR9: Yeah, but gives two to ease the pain slightly

kubejs/server_scripts/create_extra_gauges/recipes.js
```diff
@@ -0,0 +1,161 @@
+// priority: 0
+"use strict";
+
+const registerCreateExtraGaugesRecipes = (event) => {
+if (Platform.isLoaded("extra_gauges")) {
```
Mqrius: Please indent the whole `if` for consistency with the rest of the codebase, otherwise it'll happen in a random other commit like happened [here](https://github.com/TerraFirmaGreg-Team/Modpack-Modern/commit/114365a6f1c0d31b0f4fe4bc802e0400d097fad6#diff-01a59d9818b863e10dedc1f1b633f119956264176957aff2a3b4ef86dd6b0c60) which makes that commit's changes unclear

Mqrius (approved):
Code-wise looks good, gameplay-wise no idea but it's optional so probably fine!

PR #3163, opened 2026-02-23
Add Ceramic Circuit board

TomPlop:
missing lang file


kubejs/startup_scripts/tfg/items.js
Pyritie, line 1: are these supposed to be commented out?

TomPlop: I started by doing a custom item then Redeix said it could be simpler to just change the name and the model of the actual plastic board. Which made sense. I kept the comment out just in case.

Pyritie: ah yeah sounds good

PR #3188, opened 2026-02-25
Fixed exception spam

Nebby1999:
## What is the new behavior?
Fixes unticketed bug where the Sculptor+ tooltip would throw exceptions multiple times, reducing framerate

## Implementation Details
Just dont go out of bounds if index is out of bounds

## Outcome
yknow, its fixed.

## Additional Information

image

Meow


Pyritie:
thanks <3

PR #3189, opened 2026-02-25
Floaty drum tags

Mqrius:
Goes with https://github.com/TerraFirmaGreg-Team/Core-Modern/pull/319

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 ### Changes
 ### Bug fixes
+- Fixed multiple exceptions being thrown during tooltip creation (#3188) @Nebby1999
 ### Translation updates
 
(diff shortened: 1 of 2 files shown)


Redeix (approved):
Stuff like this you could always just commit directly 
:P

Redeix:
Hmm actually. Shouldn't the food oils be added too? Like seed oil and olive oil?

Mqrius:
Nah those are in #firmalife:oils, I don't think we ever need both in one tag except in this one specific use case. The mixin checks both the firmalife tag and this one.

Redeix:
> Nah those are in #firmalife:oils, I don't think we ever need both in one tag except in this one specific use case. The mixin checks both the firmalife tag and this one.

Probably true

PR #3210, opened 2026-02-27
globe crafting recipes / KubeJS

nixiee0:
# I decided to add recipes for making globes so they could be used for decoration. Crafts are based on the planet's stones. 
## So it’s like an achievement for visiting the planet)

image


Pyritie:
thanks but we already have plans for these globes, we're just waiting until more of the modpack has "settled down"

PR #3216, opened 2026-02-28
Add Artisan Table recipes and MV Chapter rework

TomPlop:
Don't forget to grab new core

### DIFF
--- a/kubejs/startup_scripts/tfg/moon/materials.moon.js
+++ b/kubejs/startup_scripts/tfg/moon/materials.moon.js
@@ -70,7 +70,7 @@ function registerTFGMoonMaterials(event) {
 
 	event.create('armalcolite')
 		.gem(2)
-		.components('1x magnesium', '1x iron', '1x ilmenite')
+		.components('1x magnesium', '1x ilmenite', '2x oxygen')
 		.color(0x443333)
 		.secondaryColor(0x5e2c21)
 		.iconSet(GTMaterialIconSet.getByName('tfc_emerald'))
@@ -83,10 +83,10 @@ function registerTFGMoonMaterials(event) {
 		.secondaryColor(0xF35A6C)
 		.iconSet(GTMaterialIconSet.getByName('tfc_hematite'))
 		.ore()
-		.addOreByproducts('olivine', 'ilmenite', 'ilmenite')
+		.addOreByproducts('olivine', 'ilmenite', 'rhenium')
 		.ingot()
 		.liquid()
-		.blastTemp(2800, $BlastProperty.GasTier.MID, GTValues.VA[GTValues.HV])
+		.blastTemp(2212, $BlastProperty.GasTier.MID, GTValues.VA[GTValues.HV])
 		.flags(
 			GTMaterialFlags.FORCE_GENERATE_BLOCK, 
 			GTMaterialFlags.GENERATE_PLATE,


kubejs/server_scripts/tfg/hv_rework/material.recipes.js
```diff
@@ -0,0 +1,55 @@
...
+
+	//#region Modify Electronic Component to require Silicon Rubber
+
+	event.replaceInput({ id: 'gtceu:shaped/filter_casing' }, Item.of('gtceu:steel_frame'), Item.of('tfg:mo_50_re_frame'))
+	event.remove({ id: 'gtceu:assembler/plascrete'})
+	event.remove({ id: 'gtceu:assembler/cleanroom_glass'})
+
+	event.recipes.gtceu.assembler('tfg:plascrete')
+		.itemInputs(
+			Item.of('tfg:mo_50_re_frame', 1),
```
Pyritie, line 20: I think we should still use relatively simple materials for the cleanroom, especially since we want to encourage making a big one from the start -- making it too expensive will just get people to cheap out again.

kubejs/server_scripts/tfg/hv_rework/material.recipes.js
```diff
@@ -0,0 +1,55 @@
...
+			Item.of('gtceu:polyethylene_plate', 6),
+			Item.of('gtceu:steel_plate', 4))
+		.inputFluids(Fluid.of('gtceu:polyethylene', 144))
+		.itemOutputs(Item.of('gtceu:plascrete', 2))
+		.duration(20 * 10)
```
Pyritie, line 25: maybe the duration should be shortened too so players can quickly make a lot of it

kubejs/server_scripts/tfg/hv_rework/material.recipes.js
```diff
@@ -0,0 +1,55 @@
...
+			Item.of('gtceu:polyethylene_plate', 6),
+			Item.of('gtceu:steel_plate', 4))
+		.inputFluids(Fluid.of('gtceu:glass', 144))
+		.itemOutputs(Item.of('gtceu:cleanroom_glass', 2))
+		.duration(20 * 10)
+		.EUt(GTValues.VA[GTValues.MV])
+
+	// Allow HV Cutter at HV (Ultimet)
+
+	event.recipes.gtceu.chemical_bath('tfg:ultimet_cool_down')
```
Pyritie, line 40: were we not gonna change to a different material instead of ultimet? what about using your mo-50-re here instead?

kubejs/startup_scripts/tfg/moon/materials.moon.js
```diff
@@ -70,26 +70,27 @@ function registerTFGMoonMaterials(event) {
 
 	event.create('armalcolite')
 		.gem(2)
-		.components('1x magnesium', '1x rutile', '1x ilmenite')
+		.components('1x magnesium', '1x iron', '1x ilmenite')
```
Pyritie: for chemical accuracy, change this to 1 magnesium, 1 ilmenite, and 2 oxygen
https://en.wikipedia.org/wiki/Armalcolite

kubejs/startup_scripts/tfg/materials/materials.coils.js
```diff
@@ -34,8 +34,22 @@ function registerTFGCoilMaterials(event) {
...
-		.flags(GTMaterialFlags.DISABLE_DECOMPOSITION,
-			GTMaterialFlags.GENERATE_PLATE)
+		.flags(
+			GTMaterialFlags.DISABLE_DECOMPOSITION,
+			GTMaterialFlags.GENERATE_PLATE,
+		)
+		
+	// Nichrome
+
+	event.create('tfg:desh_tempered')
```
Pyritie, line 44: if this is for the coils, I don't need it

kubejs/server_scripts/tfg/hv_rework/material.recipes.js
Pyritie: maybe we could use something from the stone dust line instead?

kubejs/server_scripts/tfg/hv_rework/material.recipes.js
TomPlop: Let's not make stuffs actually easier that's the total opposite of what we should do. Let's remember right now you can do HV in hours.

kubejs/server_scripts/tfg/hv_rework/material.recipes.js
TomPlop: That's actually cheap and easy. Also Mo-50-Re is a great material for here and an other really big incentive to use ore processing. 

If you want to add a stone dust line stuffs to it I'm not against but I think this is still a great change.

kubejs/server_scripts/tfg/hv_rework/material.recipes.js
TomPlop: Why?

kubejs/startup_scripts/tfg/materials/materials.coils.js
TomPlop: Well yeah I think we may use it in different forms so sounds better to do it that way rather that just an item. Also I will do the coils don't worry.

kubejs/server_scripts/tfg/hv_rework/material.recipes.js
Pyritie: strongly disagree. You designed Mo-50-Re as an optional line for funny rotors, not as something mandatory for one tier later. I'm fine with including it in sparingly in other crafts, but requiring it for something we want people to make big like a cleanroom is a bad idea. People already put off the cleanroom as long as they can, or make it tiny, and making it more expensive is going to have the opposite effect you want

kubejs/server_scripts/tfg/hv_rework/material.recipes.js
Pyritie: standing around waiting for a machine is not a good way to make a tier take longer. I don't expect anyone to make duplicate assemblers until post ae2 (and ae2 requires the cleanroom)

kubejs/server_scripts/tfg/hv_rework/material.recipes.js
Pyritie: because every other tier has a different material for the cutter blade. Why reuse the same one here?

kubejs/startup_scripts/tfg/materials/materials.coils.js
Pyritie: No, I'm doing the coils

TomPlop: Wait are you changing the concept we had. What are you planning exactly why do you want to do it?

kubejs/server_scripts/tfg/hv_rework/material.recipes.js
TomPlop: We said we change EV

kubejs/server_scripts/tfg/hv_rework/material.recipes.js
TomPlop: Pyritie that's Gregtech. I dont want to argue but it's insane to want to actually make recipes easier.

kubejs/startup_scripts/tfg/materials/materials.coils.js
Pyritie: Because I want to contribute something? lol
it's the same as we already discussed on thursday

kubejs/server_scripts/tfg/hv_rework/material.recipes.js
Pyritie: I don't think it's insane.

kubejs/server_scripts/tfg/hv_rework/material.recipes.js
TomPlop: Oh okay I understand you are confuse ! Mo-50-Re isn't a turbine rotor alloy. The recipe is easy as hell.

kubejs/server_scripts/tfg/hv_rework/material.recipes.js
Pyritie: oh ok

kubejs/server_scripts/tfg/hv_rework/material.recipes.js
Pyritie: oh which one am I thinking of, the yellow-brown color with the loop and the funny gas

TomPlop: Yeah! I'm not that evil! I even made a short duration mixer recipe. But I agree we could also fit it somewhere in the rockets. The goal here wasn't to make it harder (though it's a bit harder) but making a sink for Rhenium and having it mandatory at HV on something totally different from BTX.

kubejs/startup_scripts/tfg/materials/materials.coils.js
TomPlop: Oh okay fair

kubejs/server_scripts/tfg/hv_rework/material.recipes.js
Pyritie: also I just noticed you removed the concrete, add it back otherwise it's not actually plascrete any more

PR #3233, opened 2026-03-01
Vintage Alcohols

lysteryale:
## What is the new behavior?
Adds further aged alcohols that has +1 level to their effects and 50% more duration. Gives lower hydration and higher intoxication. These are made from just keeping aged alcohols in sealed barrels for 24 more days, has a 25% reduction in the final yield. 

## Implementation Details
All done in kubejs. 

## Outcome
Added stronger alcohols and their bucket tooltips.

## Additional Information
No langs for the recipe and fluids yet (I'm scared).

Shouldn't have any, only use for the added fluids is for drinking. 

Discord username: broofsi


Pyritie:
you'll want to add a new page to the drinks section in the field guide too

PR #3237, opened 2026-03-02
A bit of everything HV and quests

TomPlop:


kubejs/startup_scripts/tfg/materials.js
```diff
@@ -165,6 +165,51 @@ const registerTFGMaterials = (event) => {
 			GTMaterialFlags.GENERATE_BOLT_SCREW,
 			GTMaterialFlags.GENERATE_FRAME)
 
+	event.create('tfg:molten_felsic_silicate')
```
Pyritie: why not just add a liquid to the igneous felsic material?

TomPlop: that's a great question

PR #3275, opened 2026-03-06
Fix fluid modification

TomPlop:
Recreate the recipe because replaceInput doesn't work with fluids
If a recipe is modified through the one called by function it will crash and if an input fluid isn't a tag it will crash too


Mqrius (commented):
I think #3278 is cleaner because with 3278 we can just use the old replaceInput code

PR #3278, opened 2026-03-06
Add #forge tags to fluids we want to replace

Mqrius:
## What is the new behavior?
Should fix the fluid replacement issues

Closes #3275 because it's not necessary anymore

## Implementation Details
Add the #forge tags that GT uses internally early on in the process. That way the recipe modification works because the forge tags are not empty.

## Outcome
\o/

## Additional Information
Hopefully GT will stop doing the weird tag thing, then we don't need this anymore. It won't break anything if they change it, luckily.


TomPlop:
They are already registered that way?

Mqrius:
Yes but no 😭 not at the time of recipe modification.

PR #3279, opened 2026-03-06
Fix craft gun Freedom in recipes.guns.js

Tipy4:
Fix craft gun Freedom in recipes.guns.js


Tipy4:
Previously this recipe produced a bucket from stainless steel plates. The recipe configuration has been fixed, and the gun can now be crafted.

Pyritie (approved):

PR #3282, opened 2026-03-06
Fix circuits and add modification function

TomPlop:
Adds a function to modify GTm recipes with different examples. It's a bit finicky so always test that it works before pushing.


kubejs/server_scripts/tfg/utility.recipes.js
```diff
@@ -0,0 +1,203 @@
...
+    })
+
+    // Modify item input quantity
+    global.modifyRecipe(event, "gtceu:assembler/some_recipe", {
+        itemInputs: { "gtceu:copper_plate": 4 }
+    })
+
+    // Modify fluid input quantity
+    global.modifyRecipe(event, "gtceu:chemical_reactor/some_recipe", {
+        fluidInputs: { "gtceu:sulfuric_acid": 500 }
```
Pyritie: does this one need to use a tag too?

TomPlop: yeah my mistake

any fluidInput should be a tag except if it was modified already

PR #3302, opened 2026-03-07
Fix: made separate LCR/CR aluminium sulfate recipes and sodium sulfate electrolyzer path

sqcode06:
Updated recipes for aluminum processing and sodium sulfate recycling in the `material.recipes.js` script. The changes are: updating the aluminum sulfate recipe to use a large chemical reactor, adding a new variant for the chemical reactor, and introducing a recycling process for sodium sulfate.


kubejs/server_scripts/tfg/hv_rework/material.recipes.js
```diff
@@ -110,6 +110,15 @@ function registerTFGHVMaterialRecipes(event) {
 		.EUt(GTValues.VA[GTValues.LV])
 		.circuit(2)
 
+	event.recipes.gtceu.chemical_reactor('tfg:aluminium_sulfate')
```
Pyritie: does this one need its auto-generated LCR recipe removed?

kubejs/server_scripts/tfg/hv_rework/material.recipes.js
```diff
@@ -123,6 +132,13 @@ function registerTFGHVMaterialRecipes(event) {
 		.itemOutputs(Item.of('tfg:alumina_dust', 1))
 		.duration(20 * 8)
 		.EUt(GTValues.VA[GTValues.HV])
+	
+	event.recipes.gtceu.electrolyzer('tfg:sodium_sulfate_recycling')
```
Pyritie: it's better to add the electrolyzer flag to the material definition instead of making your own electrolyzer recipe

kubejs/server_scripts/tfg/hv_rework/material.recipes.js
sqcode06: I mean, in my world it works fine as is. You can check yourself to be sure

kubejs/server_scripts/tfg/hv_rework/material.recipes.js
Pyritie: https://github.com/TerraFirmaGreg-Team/Modpack-Modern/blob/dev/kubejs/startup_scripts/tfg/materials/materials.alumina.js?plain=1#L30C1-L30C42 here

Pyritie:
I'll let tom decide if this has any side effects or not, lol. We usually disable a lot of decomp on materials if the material is like, part of a recipe chain we don't want people to skip early out of or something

kubejs/server_scripts/tfg/hv_rework/material.recipes.js
TomPlop: Could just remove the CR one

Pyritie: like this?

TomPlop:
I was wondering if we really want to allow decomposition but I guess an other avenue for sulfur isn't that bad.

sqcode06:
The thing is, the sodium sulphate dust is useless. Literally useless. This way it could have at least some value, at least until another way later.

TomPlop:
> The thing is, the sodium sulphate dust is useless. Literally useless. This way it could have at least some value, at least until another way later.

This is something I utterly don't give a flying fuck about because the modpack is in Beta. I can't give utility to everything right from the start. So balance or gameplay will never be sacrificed for giving uses to "stuffs". In that case I don't really mind that you get an infinite Sulfur though I think it's kinda bad because Venus should have been the one. But I will see and could remove the electrolyzation again later.

Pyritie:
I mean you can already get infinite sulfur from a tree farm

Issue #3321, opened 2026-03-08
[💡]: upgraded spindles


gayleywilliams:
### Was this feature suggested before?

- [x] No, it was not.

### A list of popular features/mods the team has already considered and will not implement.

- [x] I understand :(

### What would you like to be added?

the ability to upgrade to a higher tier of spindle

### Why is this needed?

ceramic spindles have a durability that doesn't keep up and i think it would be nice to create a more resilient spindle for metallurgy ages such as in this mod: [Spindles TFC](https://www.curseforge.com/minecraft/mc-mods/spindles-tfc)

if not, no worries, just a little QoL thing i think would be useful

thank y'all for all your work on this pack. this was my first time playing minecraft at all and playing this has made it nearly impossible to get into any other packs 😅

Pyritie:
I'm fine with an upgraded spindle, but I don't think we need a whole extra mod for it. Just one metallic one would be fine imo, since you'll eventually be able to use a wiremill/coiling machine to get string too.

DrEthan77:
Ive regained confidence, I'll claim this one.

PR #3345, opened 2026-03-09
Spindle Rework

DrEthan77:
## What is the new behavior?
Redoes Spindle recipes to allow for anything tagged #tfg:tools/spindles

## Implementation Details
Ive gotten everything except how to add a tag to fluid copper and how to actually make the casting recipe work. 

## Outcome
Addresses #3321 

Any mod's spindle recipes cannot be bulk updated with KubeJS, so they have to be removed and re-added manually.


kubejs/server_scripts/tfg/primitive/recipies.spindle.js
```diff
@@ -0,0 +1,23 @@
+"use strict";
+
+function registerTFGSpindleRecipies(event) {
+	event.recipes.tfc.damage_inputs_shapeless_crafting(
```
Pyritie: would be good to put all of the spindle recipes in one place (maybe these could all go in the cloth file?)

DrEthan77:
I mostly did the mold for the off chance that different metals are added in the future, like bronze.

PR #3351, opened 2026-03-10
Fix 3 Deafission recipe loading errors (missing duration)

Tipy4:
## What is the new behavior?
This PR fixes 3 Deafission recipes that failed to load because the required `duration` field was missing.

## Implementation Details
- Added the missing `duration` value to 3 Deafission recipes.

## Outcome
- The `No key duration` recipe load errors are removed.
- All 3 affected Deafission recipes now load correctly.

Low risk. Changes are limited to recipe timing fields (`duration`) for 3 Deafission recipes.


Pyritie:
we're going to talk to the mod author for a better solution

PR #3390, opened 2026-03-12
Fix jam sandwich food processor recipes

### LINKED ISSUE #3388: [🐛]: Jam sandwhiches missing recipe tag

- [x] No, it was not.

### Computer OS

Windows 11

### Modpack Version Used

0.11.26

### Minecraft Environment

Forge Server

### New Worlds

Yes

N/A

### Expected Behavior

Being able to craft jam sandwhiches inside a food processor using most types of protein and dairy, as it is currently possible with a regular crafting recipe.

Image

### Actual Behavior

Currently jam sandwiches are only able to be made inside the food processor with either only jam or moon cheese on slices of "cheese". 
Image

### Steps to Reproduce

Get a food processor or check EMI to verify that the regular recipes are not present in the food processor

### Additional Information

This is likely caused because the food processor jam sandwhich recipe uses only the tag tfc:foods/usable_in_jam_sandwhich_2 and not also tfc:foods/usable_in_jam_sandwhich, dont know if its intended beahviour or not.

Mqrius:
## What is the new behavior?
Fixes #3388

## Bug details
The useable_in_jam_sandwiches_2 tag got built from the other tags too early, before all items were registered. So it ended up not having the required dairy.

## Latent bug details
GT doesn't support having recipes with the same input items but different amounts, so you can't have [1 jam 2 cheese] and [2 jam 1 cheese] as distinct recipes.

## Fix details
Remove useable_in_jam_sandwiches_2
Make a copy of the preserves tag called tfc:foods/preserves_2
Use preserves_2 for the [2 jam 1 cheese] sandwich to give it distinct inputs for GT's RecipeDB

## Discarded fix that would work also work
Don't build useable_in_jam_sandwiches_2 from the other tags by iterating, instead define it lazily out of other tags (#cheese, #preserves, etc) that will have the correct items in the future. You'd still need to make the recipes distinct by using _2 only for one of them but not the other.

### DIFF
--- a/kubejs/server_scripts/tfg/food/recipes.food.js
+++ b/kubejs/server_scripts/tfg/food/recipes.food.js
@@ -199,11 +199,11 @@ function registerTFGFoodRecipes(event) {
 				])
 			})
 
-			//Note: Jam needs to be first in the recipe code or else it will consider it as the usable_in_jam_sandwhich ingredients.
-			//1 Jam
+			//Note: preserves needs to be first in the recipe code or else it will consider it as the usable_in_jam_sandwich ingredients.
+			// 1 jam + 2 cheese
 			global.processorRecipe(event, `${grain}_${type[0]}_jam_sandwich_1`, 100, 16, {
 				circuit: 4,
-				itemInputs: [`2x ${type[1]}`, '#tfc:foods/preserves', '2x #tfc:foods/usable_in_jam_sandwich_2'],
+				itemInputs: [`2x ${type[1]}`, '#tfc:foods/preserves', '2x #tfc:foods/usable_in_jam_sandwich'],
 				itemOutputs: [`2x tfc:food/${grain}_bread_jam_sandwich`, 'tfc:empty_jar'],
 				itemOutputProvider: TFC.isp.of(`2x tfc:food/${grain}_bread_jam_sandwich`).meal(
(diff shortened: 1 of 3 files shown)


Redeix:
Couldn't you cut all recipes in half by doing 1 jam + 1 cheese + 1 (jam + cheese)

Mqrius:
In theory yes, but it comes with the drawback that we have to remember to add food to extra tags when we add them. TFC doesn't offer `usable_in_jam_sandwich_except_jam`, so for special stuff like moon cheese and 'slice of cheese' you'd have to add it to two different tags.

Even if you do, it only removes a third, because your proposal doesn't include 3 jam.

Mqrius:
If you want less recipes, you're better off collapsing the bread types imo

Redeix:
No, you would just make a new tag that combines the cheese tag and the jam tag. Anything that gets added to those tags in the future will automatically be added into the third new tag.

Edit: actually yeah, it's like the same amount of work either way since there are exceptions.

Redeix (approved):

Mqrius:
I'm using "cheese" as a shorthand for stuff that goes in jam sandwiches that's not jam: cheese, meat, egg, 'slice of cheese', and fish roe
image

Redeix:
Yeah I was just confusing myself about that. My b

PR #3417, opened 2026-03-14
Vitrified ender pearls use ender pearl tag

Bumperdo09:
## What is the new behavior?
Vitrified ender pearls use ender pearl tag

## Outcome
So you can use the ender pearls dropped by beneath endermen


Pyritie:
this is intentional

PR #3422, opened 2026-03-14
Enhanced Visuals Alternative Rain Effect

applenper:
it is no longer blue and ugly

this solves the issue of the rain effect being blue and ugly


Redeix (dismissed):
Just to double check but this rain drop texture is free use right?

applenper:
to my understanding yes, i sourced it from https://clipart-library.com/ which allows use of its images as long as they dont constitute commercial use

config/enhancedvisuals-client.json
```diff
@@ -1,4 +1,4 @@
-{
+F{
```
Pyritie: what is the `F` up here? it looks like a typo

Redeix: oop

Redeix (approved):

config/enhancedvisuals-client.json
applenper: it's GONE NOW

Pyritie (approved):

PR #3431, opened 2026-03-15
Allow tfc bricks to stonecut into other brick variants

SpicyNoodle5:
## What is the new behavior?
Re-adds recipes for regular bricks -> all of its stonecutter variants (e.g marble bricks -> square marble bricks, andesite bricks -> create brick variants, etc)

## Implementation Details
For each TFC rock, just add a recipe to use the brick for each given stonecutter variant. Maybe not the cleanest solution, but it works.


Pyritie:
does this problem only happen for the tfc bricks? what about the non-tfc bricks?

I'd prefer if you used the rock table's `rock.bricks.block` instead of the string, or whatever the ID is sorry I'm on my phone

SpicyNoodle5:
Changed and yes it was the same for other bricks. I did notice for some of the non-tfc rocks, some of the recipes were duplicated but that seemed to be pre-existing so maybe its fine?

Pyritie:
for the stonecutting? yeah, it's a known issue, but I guess a similar `.isTFC` check could be done instead to avoid that problem?

Issue #3419, opened 2026-03-14
[💎]: Add Alternative To Chorus Fruit For Nichrome Coils


dos54:
### Was this enhancement suggested before?

- [x] No, it was not.

### What would you like to be added/ changed?

Currently, Fired Chorus-Infused Ceramic is required to make Nichrome Coils. As part of the craft the player is required to use chorus fruit and popped chorus fruit... which is a food. There should either be an alternative crafting recipe for nichrome coils that does not involve using a food for the craft, or chorus fruit should be changed to no longer be a "food" (and thus have the extra data applied by TerraFirmaCraft removed)

An appropriate alternative would be to allow either popped chorus fruit, OR chorus flower to be used in the recipe.

### Why is this needed?

AE2 does not interact well with TFC's food system. While it's not much of an issue with most food items (you're probably better off just setting up a passive system for food production), it IS an issue with more complex autocrafts such as nichrome coils. Workarounds exist, but they're janky and likely to break and, at the end of the day, just aren't fun. Adding an alternative recipe that does not include a food item (such as using chorus flowers instead of fruit) would fix this issue without breaking the intended moon-gate.

Pyritie:
can you make the recipe ignore nbt?

dos54:
> can you make the recipe ignore nbt?

Pattern recipes do not allow ignore NBT or substitutions, and the pattern provider does not allow the fuzzy card. The workaround I mentioned before is using a fuzzy card to keep an input bus stocked on the machine always, but it's janky, can easily break, and makes patterning inconsistent.

edit: processing recipes*, not pattern recipes

Pyritie:
hmm ok

TomPlop:
But why not making it a passive?

dos54:
> But why not making it a passive?

Passive nichrome coils? 🤮 you have to fight to get me to passive hydrogen lol

Pyritie:
no, passive the chorus fruit -- it's the only real way to get nitrogen on the moon

TomPlop:
yeah just have the insulation passively and you good because then it doesn't rot. If you are smart you can also have on demand Chorus Infused Ceramic simply by removing the chorus from the pattern and having it passively in your implosion compressor on the moon.

All fo that is really good to make you think and to make you use Interplanetary logistics so I don't see why we should change it. I would be more concern if it was a final item you want to store.

dos54:
[...]

PR #3451, opened 2026-03-17
Granulated Chorus Dust

AirRice:
## What is the new behavior?
Adds new items Granulated Chorus Dust and Purpur Brick, which should solve some of the hassle of trying to make nichrome coils & purpur bricks from an expirable resource (popped chorus fruit)

## Implementation Details
Since Popped Chorus Fruit is edible & expirable in TFG, using them in automated crafting can be difficult or tedious (#3419) especially when interfacing with AE2. This PR fixes that by adding a new intermediary item, which is then used in nichrome coil recipes (Bottom left for new material):
image

This can be obtained by macerating either popped chorus fruit or chorus flowers.
image

Additionally this granulated chorus dust can also be turned into brick form in an alloy smelter, before it's then used to make purpur bricks (changed from current recipe which uses popped chorus)
image
image

## Outcome
Addresses #3419.
Makes Purpur bricks also more easily crafted without worrying about the issues mentioned above
Also fixes some weirdness from letting purpur bricks be made directly from popped chorus, like popped chorus having the 'tfg:stone_brick' tag

## Additional Information
I'm still not sure about letting people get this from chorus flowers, maybe we should force this only to be popped chorus to prevent people from skipping the process of setting up the EGH on the moon entirely.

N/A


TomPlop:
Let it be known that I dislike this idea. The chorus fruit presented many interesting choices from having a passive system on the moon to grow it constantly. To use voiding cover or make pattern with a different methods. 

This PR homogenize it to make it one more boring pattern. The issue wasn't an issue. I would also like to remember that Redeix and Pyritie wanted to make Canola a food too while there wasn't any void cover for food at that time and it could easily clock inventory.

But for this one that is easily solvable and present an interesting challenge to the player. We should solve it by making it boring? That's very weird.

Pyritie:
While I do agree that we need some automation challenges that you can't just pattern away, I don't think chorus fruits are really the right place for it. When I made them part of the new coils, my intent was to "check" that players had a moon base going, and chorus seemed like a good place since they were already a big part of it.
Making it easy for players to upgrade all their machines with new coils is something we're already wanting to do anyway, isn't it?
Maybe we could have the powdered chorus be part of the moon "line" anyway. So as part of processing chorus for nitrogen, you also get the chorus dust as a byproduct. That'll encourage passiving the nitrogen. Maybe the powdered chorus could be used as a food additive too then, sorta like wraptor sugar lol

AirRice:
After discussion in the devcord I'm closing this one since the actual issue presented is not as much of an issue as I thought initially

Issue #3437, opened 2026-03-16
[💡]: Recycle Hatch/Bus


Xisengan:
### Was this feature suggested before?

- [x] No, it was not.

### A list of popular features/mods the team has already considered and will not implement.

- [x] I understand :(

### What would you like to be added?

It seems to be impossible to recycle the Item/Fluid Hatch/Bus (LV, MV, HV...). I mean, recycle them to their materials, like you would do with machines. Would it be possible to add a way to recycle them?

### Why is this needed?

So you don't have to delete the item. I get it that these can be re-used, but I have a lot of them spare and would like to recycle some.
Also, you can recycle the energy hatches (and I get it, they do get outdated and can't be re-used in higher tiers). If it's not a problem for you guys, I'd love the recycle option. I thank you a lot for your hard work. Have a great day.

Pyritie:
we need to be careful with the MV ones since the hatch won't remember which bronze was used to create them

PR #3470, opened 2026-03-19
Added recycling recipes for GT Fluid/Item Hatch/Bus. LV -> IV

nixiee0:
- https://github.com/TerraFirmaGreg-Team/Modpack-Modern/issues/3437


SpicyNoodle5:
Ik this has already been merged, but do we still need to be careful about MV hatches recycling into bronze? (They are made with any bronze drum)

nixieeq:
Be careful—recycling yields regular bronze

> Ik this has already been merged, but do we still need to be careful about MV hatches recycling into bronze? (They are made with any bronze drum)

Pyritie:
good point, fixed

Issue #3456, opened 2026-03-18
[💎]: Register GT fuels as valid train fuels


Pyritie:
### Was this enhancement suggested before?

- [x] No, it was not.

### What would you like to be added/ changed?

Add things like diesel etc to the fuels that trains are able to consume in order to go fast

https://github.com/Layers-of-Railways/Railway/wiki/Registering-a-fluid-as-a-Liquid-Fuel-Type

### Why is this needed?

Train fuel tanks are useless right now

SakuraKitsurugi:
Can look into this

PR #3473, opened 2026-03-19
SNR Fuels

SakuraKitsurugi:
### What is the new behavior?
SNR can now use most of the combustion fuels as well as creosote and biomass. Lava is also buffed to be more in-line.

| Fuel  | t/100mB |
| ------------- | ------------- |
| Bio Diesel | 3200  |
| Diesel | 4800  |
| Ethanol | 2400 |
| Toluene | 4000 |
| Octane | 1000 |
| Gasoline | 20000 |
| Rocket Fuel | 3125 |
| HOG | 40000 |
| Naphtha | 800 |
| CBD | 12800 |
| Methanol | 800 |
| Lava | 50 => 900  |
| Creosote | 600 |
| Biomass | 850 |

### Implementation Details
Followed the liquid fuel implementation guide and added datapack files.

### Additional Information
- Weaker fuels mostly include Oils which I did not add due to not really being worth it. (Also to just in case prevent it from using it if the Oil is being transported)
- #3456 

### Discord
@sakura.kitsurugi

### DIFF
--- a/kubejs/data/railways/railways_liquid_fuel/creosote.json
+++ b/kubejs/data/railways/railways_liquid_fuel/creosote.json
@@ -1,0 +1,6 @@
+{
+  "fluids": [
+    "gtceu:creosote"
+  ],
+  "fuel_ticks": 600
+}


Pyritie:
what do we think about adding creosote?

Pyritie:
or some of the other burnables like seed oils

SakuraKitsurugi:
Creosote does not sound bad, but seed oils going off of Large Boiler fuels is VERY weak. With a similar 4x buff it is around 125t / 100mB

TomPlop:
Isn't rocket fuel a bit low. I feel like it should be the best one 😂

SakuraKitsurugi:
Based on Combustions Generators, it is not 😂

SakuraKitsurugi:
@Pyritie by checking regular Boiler I can make Creo as the worst fuel which would be ~600t/100mB (technically 560, but meh)

TomPlop:
> Based on Combustions Generators, it is not 😂

Yeah but isn't Rocket Fuel the traveling fuel :p
I don't just feel like it would be funny to have it goated for trains lol

PR #3486, opened 2026-03-19
Make redstone etching fluid consistent with GT dusts

SpicyNoodle5:
## What is the new behavior?
Purely for recipe consistency with other GT dusts at the cost of making this use 10x more redstone. If the balancing consequences are too severe, feel free to close. This would only really matter if you could decomp these items/fluids or if they are ever used outside of this loop


TomPlop:
Sounds good to me? @Pyritie

PR #3500, opened 2026-03-21
Remove the Heat Exchanger

TomPlop:


config/ftbquests/quests/chapters/mv__medium_voltage.snbt
Pyritie, line 1: did you mean to delete these pages?

TomPlop: yeah it was the list of boosters but there is now a emi tab

PR #3501, opened 2026-03-21
Added All Missing Gem Slurries

SakuraKitsurugi:
### What is the new behavior?
If dangerous why so tasty looking.

### Implementation Details
As discussed on Discord.

### Additional Information
@TomPlop


Xtrial-01:
> If dangerous why so tasty looking.

Uranium be like:

JeanRdSz:
shouldn't Opal also have centrifuging outputs?

SakuraKitsurugi:
damn, ur right, i added it to materials but forgot centrifuge brb

PR #3528, opened 2026-03-23
Distant Horizons: set enableAdaptiveTransferSpeed to true

coolsa:
Me and my pals are really enjoying the modpack so far (very early game, copper age), but we noticed the config for distant horizons has a minor issue. This pull request is the change needed to resolve that issue.

## What is the new behavior?
This fixes an issue with really high pings, with `/ping` sometimes exceeding 20000ms.

## Implementation Details
Changed `enableAdaptiveTransferSpeed` from `false` to `true` in `DistantHorizons.toml`

## Outcome
The above setting would sometimes unintentionally exceed bandwidth limits and would act as an unintentional DDOS to any clients with Distant Horizons. Setting this to true makes it so the server now respects connect clients bandwidth limits.

This eliminates an issue (unreported) that network connectivity between clients and the server would initially be stable, but at some point packets would be increasingly delayed, and eventually players would time out and be kicked. 

I am @coolsa


Mqrius:
It sounds good but they disabled it by default on purpose. Any idea why? I can't find much info about it. https://gitlab.com/distant-horizons-team/distant-horizons-core/-/commit/b3ebaffa853c9d969efe3db95f7d9d5a4146ae64

coolsa:
> It sounds good but they disabled it by default on purpose. Any idea why? I can't find much info about it. https://gitlab.com/distant-horizons-team/distant-horizons-core/-/commit/b3ebaffa853c9d969efe3db95f7d9d5a4146ae64

Looks like it was disabled because the implementation was "a bit hacky", but also because there was some log spam. https://gitlab.com/distant-horizons-team/distant-horizons/-/work_items/1090#note_2652633705

I looked over the code that implemented it, and it looks good to me. https://gitlab.com/distant-horizons-team/distant-horizons-core/-/commit/3ca5efadc9c19c607b387795d3dbe25214fb0ea2

My server has been up for roughly 5 days at this point without any crashes or restarts. I'm pretty confident that it's a safe change, but someone else should test it to confirm that there are no unforeseen consequences or stability impacts.

Mqrius:
Fair enough. And is this a clientside setting, or serverside, or both? Cos if it's only serverside then we have a specific place for it.

coolsa:
> Fair enough. And is this a clientside setting, or serverside, or both? Cos if it's only serverside then we have a specific place for it.

I just had a friend test it out, seems like it needs to be enabled on both the client and server.

Issue #3527, opened 2026-03-23
[🐛]:Raw Pizza not craftable?


abculatter2:
### Was this bug reported before?

- [x] No, it was not.

### Are you using Prism or Curseforge?

Yes, I am using Prism or Curseforge.

### Have you tried redownloading the pack?

- [x] Not Applicable / Yes, I have tried downloading a new instance.

### Computer OS

Windows 11

### Modpack Version Used

11.28 (though was also present in 11.27)

### Minecraft Environment

Singleplayer

### New Worlds

Yes

### Please provide a link to your latest.log, if relevant.

N/A

### Expected Behavior

Not sure if this is a bug, but I would expect the "Raw Pizza" item to be craftable in a crafting table.

### Actual Behavior

It is only craftable in a food processor.

### Steps to Reproduce

Look in EMI.

nixieeq:
I might not have understood you quite right, but here’s the craft on the workbench. 11.27

<img width="337" height="163" alt="Image" src="https://github.com/user-attachments/assets/3fc00334-ae91-4b44-b22a-b26cf3415818" />

abculatter2:
> I might not have understood you quite right, but here’s the craft on the workbench. 11.27
[...]

PR #3542, opened 2026-03-24
Workbench Recipes | Firmalife pizza

nixiee0:
As I understand it, this issue was described in this issue(https://github.com/TerraFirmaGreg-Team/Modpack-Modern/issues/3527)


Pyritie:
firmalife already has these recipes. The problem is that gurman/gourmet disables them. Let me see if there's a better way to enable the original recipes.

Issue #3520, opened 2026-03-22
[💡]: Metal Girder Crafting Amount


flatpop:
### Was this feature suggested before?

- [x] No, it was not.

### A list of popular features/mods the team has already considered and will not implement.

- [x] I understand :(

### What would you like to be added?

Metal girders from Create currently feel too expensive for a primarily decorative block. I propose adjusting their crafting output OR adding tiered recipes.

Proposed solutions:

Increase the recipe output to 8 girders

OR

Add multiple recipe tiers based on material:
Wrought iron: 4 girders
Steel: 8 girders
Aluminum: 16 girders
Stainless steel: 32 girders
Titanium: 64 girders
(this may seem excessive, but these are decorational blocks used in large quantities, and the only actual use they have is for multiblocks that will be long outclassed by the time you get stainless steel and titanium)

### Why is this needed?

Metal girders are mostly decorative and are typically used in large quantities when building. The current cost (3 wrought iron plates and 3 screws for only 3 girders) makes them prohibitively expensive for this purpose.

Even in functional builds like steam pressers and liquefaction towers, girders are already the cheapest component, so adding these tiered recipes would not meaningfully affect game balance

Pyritie:
note that these are currently used to make monorail tracks as well

SakuraKitsurugi:
I can pick this up, I'll make it a const list so if the values are too strong it can be adjusted.

Pyritie:
I think just increasing the amount per craft would be fine, they don't need to be super fancy

SakuraKitsurugi:
Ah yeah, its just a const for material and amount given per recipe

SakuraKitsurugi:
ah, mb, didnt get what u mean at first, but making at least 3 different types giving a bit more should be fine?

Pyritie:
I just don't think it needs to be all that complicated lol

flatpop:
I think having ones for just wrought iron and steel would be fine, I just added the others because I love when higher tier items can be used in lower tier recipes for more output

edit: i looked at the implementation SakuraKitsurugi made and I really like it

PR #3549, opened 2026-03-24
Girder costs based on material

SakuraKitsurugi:
### What is the new behavior?
Girders have 3 new recipes that use different materials and give different yields.

### Implementation Details
- Added 3 recipes that use Cast Iron/Wrought Iron/Steel and give 4/8/16 girders.
- Added an assembler recipe to make girders.
- Adjusted the monorail yield from 8 to 6 to be in line with new costs.

## Additional Information
- #3520

### DIFF
--- a/kubejs/server_scripts/create/recipes.js
+++ b/kubejs/server_scripts/create/recipes.js
@@ -1324,7 +1324,8 @@ const registerCreateRecipes = (event) => {
 		.itemInputs(`3x #forge:plates/${material.metal}`, `3x #forge:bolts/${material.metal}`)
 		.itemOutputs(`${material.amount}x create:metal_girder`)
 		.duration(100)
-		.EUt(4)
+		.EUt(20)
+		.circuit(6)
 	})
 
 	// Стеклянная дверь


kubejs/server_scripts/create/recipes.js
```diff
@@ -1305,13 +1305,28 @@ const registerCreateRecipes = (event) => {
...
+			A: `#forge:plates/${material.metal}`,
+			B: `#forge:bolts/${material.metal}`
+		}).id(`tfg:create/shaped/metal_girder_from_${material.metal}`)
+
+		event.recipes.gtceu.assembler(`tfg:create/metal_girder_from_${material.metal}`)
+		.itemInputs(`3x #forge:plates/${material.metal}`, `3x #forge:bolts/${material.metal}`)
+		.itemOutputs(`${material.amount}x create:metal_girder`)
+		.duration(100)
+		.EUt(20)
+		.circuit(6)
```
Pyritie, line 1328: is this circuit free? I know a load of different assembler recipes want cast/wrought/steel plates

SakuraKitsurugi: technically it can be crafted without a circuit at all, but I added it just in case

SakuraKitsurugi: (no conflicts if you put only those two items in assembler is what i mean)

SakuraKitsurugi:
I'm [nearly] always on speed dial haha

SakuraKitsurugi:
Went through circuit 6 recipes and there are definitely no conflicts. The only one that is close uses 6 steel and 1 steel frame, but that shouldn't really conflict (you need bolts for girders).
image

PR #3554, opened 2026-03-25
Tier Locked Blocks Lang File

DrEthan77:
## What is the new behavior?
Adds EN lang files for the tier locked block's Jade plugin.


Pyritie:
put lang strings over here instead please https://github.com/TerraFirmaGreg-Team/Tools-Modern/tree/dev/LanguageMerger/LanguageFiles/tfg/en_us

PR #3574, opened 2026-03-26
Wood Gear Assembler Recipes

### LINKED ISSUE #3550: [💎]: Automation/integration of wood gears with gregtech

- [x] No, it was not.

Add a small wood gear and wood gear recipes to the gregtech assemblers, possibly using liquid glue instead of the glue ball item

### Why is this needed?

Right now if you want to use the cheapest kind of gears for crafts you need to manually craft wood gears with a saw, else you need to spend iron/steel for the higher tier. This impacts mainly greate basic saws, which players may want to make in bulk for big tree farms, chain conveyors, windmill bearings, mechanical bearings and a lot more. 
This improvement would make create logistics and contraption more accessible without changing the actual cost, just making the more cost effective way more accessible once you reach the assembler.
In tune with this I would propose to use liquid glue so that you don't need to convert the sticky resin to the glue item much like many create/greate recipes (see mechanical bearing, frogport ect.ect.)

SakuraKitsurugi:
### What is the new behavior?
Adds small and normal wood cog recipes to the assembler.

### Implementation Details
- Modified `kubejs/server_scripts/tfg/primitive/recipes.wood.js`

### Additional Information
- Closes #3550

### DIFF
--- a/kubejs/server_scripts/tfg/primitive/recipes.wood.js
+++ b/kubejs/server_scripts/tfg/primitive/recipes.wood.js
@@ -85,6 +85,7 @@ function registerTFGWoodRecipes(event) {
 		.itemOutputs('gtceu:small_wood_gear')
 		.duration(20)
 		.circuit(6)
+		.EUt(GTValues.VA[GTValues.LV])
 
 	event.recipes.gtceu.assembler("tfg:wood_gear")
 		.itemInputs('4x #minecraft:planks')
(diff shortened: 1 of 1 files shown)


kubejs/server_scripts/tfg/primitive/recipes.wood.js
```diff
@@ -78,4 +78,18 @@ function registerTFGWoodRecipes(event) {
 		B: '#minecraft:planks',
 		C: 'tfc:glue'
 	}).id('gtceu:shaped/gear_wood')
+
+	event.recipes.gtceu.assembler("tfg:small_wood_gear")
```
Pyritie, line 82: missing `EUt()`

SakuraKitsurugi: Sorry, fixed

Issue #3564, opened 2026-03-25
[💎]: Add Black Bronze Recipe to Alloy Smelter


abculatter2:
### Was this enhancement suggested before?

- [x] No, it was not.

### What would you like to be added/ changed?

As it is, there is a mixer recipe for electrum dust + copper dust = black bronze dust. However, there is not a corresponding recipe for this in the alloy smelter.

### Why is this needed?

Given the importance of colored steel in the super early game, it would be massively beneficial to have an alloy smelter recipe for what is, in my opinion, one of the most annoying components of the recipe. Plus, it doesn't really make much sense to not have that recipe, given the typical conventions of the pack.

abculatter2:
On a similar note, Bismuth Bronze has a similar "mixer recipe but no alloy smelter recipe" situation, as that can be made form brass + bismuth.

nixieeq:
#  I'll take it

Redeix:
There was a big argument about this one and I wasn't going to read all that to find the answer so I'm leaving it pending

Pyritie:
tl;dr we kinda want to make people use the crucible for bulk alloying some more instead of using the alloy smelter for everything, but it's gonna require some other infrastructure to work, maybe involving fixing the pump exploit and making casting channels cool through fans or something idk

PR #3576, opened 2026-03-27
black&bismuth bronze alloy smelt recipes

nixieeq:
https://github.com/TerraFirmaGreg-Team/Modpack-Modern/issues/3564 

image
image


nixieeq:
**I’ve also increased the smelting time for rose_gold and sterling_silver by 2.5 seconds. In the future, it might be worth standardizing all recipes to follow the same rule.**

> For now, I’ve set my recipes so that each ingot takes 2.5 seconds to smelt; since there are 5 ingots, that comes to 12.5 seconds.**

kubejs/server_scripts/tfg/ores_and_materials/recipes.alloys.js
```diff
@@ -314,3 +314,51 @@ function registerTFGAlloyingRecipes(event) {
...
+		"#forge:ingots/brass"
+	];
+
+	copper_types.forEach(copper_type => {
+		electrum_types.forEach(electrum_type => {
+			const id = linuxUnfucker(`${copper_type}_and_${electrum_type}`.replace(/#/g, ""));
+
+			event.recipes.gtceu.alloy_smelter(`tfg:black_bronze_from_electrum_${id}`)
+				.itemInputs(
+					Ingredient.of(copper_type).withCount(3),
```
Pyritie, line 340: I wonder if the syntax like `['gtceu:brass_dust', 'create:brass_ingot']` would work to clean up these alloy smelter recipes so we don't need all these duplicates?

TomPlop:
Not a fan progression wise they should stay mixer recipe.

It's also a way to remove some interest to the crucible. Anything that removes content from TFC is getting me on edge now lol

TomPlop:
There is a limit to how far we can push the amount of shortcut we add to remove content to TFC.

Pyritie:
Yeah but it can do regular tin bronze, which makes people gravitate towards
it more than the other bronzes

TomPlop:
Not sure I see that as an issue. Normal bronze is used for Gregtech in such big amount which isn't the case of other bronze. 

Having the mixer as the first reward to make these alloys faster is cool. Add alloy smelter that you unlock much earlier you just allow to rush even more the early game, remove interest to the mixer and remove interest to the crucible. 

Honestly I'm just tired of making TFC less that an hour of gameplay.

Pyritie:
yeah but it's just bronzes. The different bronzes should have parity in how
they're created. And there's not really much "interest" in making the mixer
just for the other bronzes. In our own group, the main appeal of the mixer
was for the colored steels and stuff like glue and compost.

Also yes you can speed through tfc but that's just because we know so much
about it. Base tfc has the same problem. I sped through tfc back in 0.7 too
lol

TomPlop:
No the different bronzes shouldn't be in parity with each other because one is used 100 times more that the other two. 

The other two are used for alloys mostly. It's not because they are all called bronze that they should be in parity. Do you want to make all the steel in parity? All the tungsten in parity? Obviously no. 

Alloy smelter unlocks much earlier. Why do they want the alloy smelter recipe? Because it allows you to not use the mixer and more important it's allow you to skip the crucible. 

I want players to use the crucible. I want them to use the anvil. I want them to play TFC. 

Now if you think it's okay to remove TFC mechanic we can go for it. But that's what you do here, you remove content from TFC.

Pyritie:
> It's not because they are all called bronze that they should be in parity.

They should be in parity because all the tfc bronzes are in parity. They all craft anvils of the same tier, have similar tools, and can be used interchangably in a lot of other recipes. That's obviously not the same as the different kinds of steels and tungstens.

You still need to use the crucible to make colored steels and to melt down the dusts you get from the mixer anyway. Unlocking easier bronzes before you unlock easier steels makes sense to me. And you can already use the alloy smelter for all the other two-ingredient pre-iron alloys like rose gold, sterling silver, brass, tin alloy, red alloy... why should these two bronzes be any different? That's the inconsistent part.

I'm not advocating for a bronze age low pressure steam alloy smelter. I'm not advocating for castable steel gears. I'm not advocating for an iron age extruder for tools.

TomPlop:
They can be used interchangeably in a lot of recipes but you will still need an insanely bigger amount of bronze. That's not even comparable.

You may not propose these now but looking at TFG, we see how insane the amount of stuffs we removed from TFC. And I'm getting tired of it. I love TFC and we just keep trimming it. 

People should use the crucible and that's it. You don't even need a lot of coloured steel which also kinda sux. 

It's not that I hate black bronze in the alloy smelter, it's just that I'm tired we remove content from TFC.

Once again if you are telling me "I want to remove usage of the crucible and I'm fine with making early game even faster" it's okay merge it.

Pyritie:
Are you suggesting we remove more recipes from the steam alloy smelter? or that we should go back to making LV machine hulls requiring 8 colored steel plates each? or removing the electric greenhouse or refrigerator? or removing the 3x3 tools? or removing the coke oven so you have to always make charcoal the tfc way? or removing any sort of permanent lighting? or removing the rock breaker so people have to quarry all the stone for their bricks? or removing the helve so you have to hammer out all of your plates and blooms? or removing ore processing so you can only melt raw ores? or removing the beneath so they can only find kaolin and graphite the tfc way? or removing any sort of backpack besides small and large vessels? or removing vanilla buckets? or removing tree farms? or removing alcohol buffs? 

they're also all "removing TFC", but I'm sure you will agree with me that being able to eventually automate a lot of the difficult problems that TFC presents is the main appeal of the modpack.

I'm not sure where you get this idea that I'm trying to remove TFC content. I love TFC as well and I'm constantly trying to add new ways to keep it relevant -- we both are. Like outside of the recent whole worldgen backport, I'm trying to find ways to make the TFC caves actually useful, and just now I made #3580 to make another basically unused TFC mechanic relevant again. I also changed the temperature sensitivity on ambiental so it's actually something you have to pay attention to and not just ignore. I would love to nerf the blue steel diving suit as well whenever GT finishes their armor system.

Pyritie:
I would also be fine with eventually removing the whole use-a-pump-to-cool-ingots-instantly mechanic because it basically makes the casting channels irrelevant. But we should only do that if we make a better solution for bulk processing, like being able to use an encased fan to make them cool faster or something.

TomPlop:
Are you suggesting we remove more recipes from the steam alloy smelter?
Which one? You already buffed the colored steel so it makes double the amount of colored steel when in the anvil it was 2 ingots for 1 ( I was against it too )

or that we should go back to making LV machine hulls requiring 8 colored steel plates each?
Let's go!!!!

removing the electric greenhouse or refrigerator?
I think we unlock the refrigerator too early and I'm fine with the EGH it's pretty balanced but we should remove Create Tree Farm and reduce drastically the amount of saplings

removing the 3x3 tools?
They are needed because GT asks for a lot more materials that TFC

removing the coke oven so you have to always make charcoal the tfc way?
Pleeeeease let's remove coal from the bloomery so people have to use the tfc way. Right now you removed the need for tfc charcoal pit ( I was against it too )

removing any sort of permanent lighting?
It's fine how it's gated and TFC does offer you blue steel and lava. We follow TFC progression here and I'm happy about it.

removing the rock breaker so people have to quarry all the stone for their bricks?
It's unlocked too early I like that it's not at least gated behind red steel

removing the helve so you have to hammer out all of your plates and blooms?
I wouldn't mind there is still the forge hammer but at least it gives option

removing ore processing so you can only melt raw ores?
Ore processing ADDS a step to melting raw ore to have better yield so it's cool. We don't remove content we add

removing the beneath so they can only find kaolin and graphite the tfc way?
I think Beneath made kaolith a bit trivial yeah

removing any sort of backpack besides small and large vessels?
I was fine with removing Sophisticated as you remembe, to have a TFC add-on

removing vanilla buckets?
It exists in TFC

removing tree farms?
Let's remove Create Tree Farm or at least let's make them saplings negative!

removing alcohol buffs?
It's a tfc add-on

Pyritie:
lol I wasn't intending to get a reply on all of those

TomPlop:
Well there is good point and I still believe trees saplings shouldnt be renewable lol

Pyritie:
lol is the coal in the bloomery really that bad? I don't see why you can't use coal for it, charcoal's appeal is that it's renewable, and then you get the coke oven as soon as you get wrought iron so it's not like it'd change much

and I'm pretty sure we'd have actual riots if we took away renewable trees, though I do agree the sapling drops are still stupidly high

aaaaanyway that aside... it kinda sounds like we could close this PR and instead maybe start thinking of ways to make the crucible relevant for alloying more, such as:
- separating steam alloy smelter recipes from the LV alloy smelter (so you can still use the steam one for molds)
- removing the pump exploit in favour of something like encased fans over casting tables

cuz I do agree it's a little weird that once you get the crucible you kinda only use it for colored steels and nothing else, when it's supposed to be a huge step up for _all_ alloying

abculatter2:
One thing I wanted to add to this as a potential way to improve the crucible and justify making it more relevant for later into the tech progression;

I feel like adding more ways to automate crucibles in the steam and early LV tier would make it a lot more interesting and rewarding to use it. As it is, the threshold switches can be used to detect molten metals in the crucible, but only if they are immediate adjacent (the inventory connector doesn't seem to be able to read fluids) and if filtered only one type of metal at a time. On the bright side they can be filtered to detect when the crucible makes a specific alloy, but that would limit each crucible to only alloying one thing each, which... Maybe that's fine?

Adding more crucible heat sources might help, too, such as being able to use the blaze burner as a heat source, or possibly a steam and/or electric forge?

It would also be really handy to have a simple "pressure plate" detector available very early on, which reads how many of the input item slots are occupied. I tried using a comparator for this but that seems to not read the crucible at all, so if there's a way to add that functionality there that would be great. I'm also on the fence on whether or not it would be good to add an earlier option for the More Red redstone circuits, as those would help a lot with making the redstone components more compact.

Overall, though, despite being (presumably) the one whose suggestion sparked this, I actually do agree with Tom on this. I felt that it was an odd omission to not have those recipes, but if more features were added that makes the crucible more viable as an alternative I would 1000% prefer that over the alloy smelter.

Also, somewhat related, would it be possible to add more of the other two bronzes as alternatives to tin bronze in some recipes? Even if it's just adding them as fluid and/or item pipes, it would be nice to have more then just tin bronze involved in steam tier crafting.

jurjen909:
I see there was already a heated discussion about this issue. But I would like to ask if it would be possible to have the alloy smelter recipe of the two coloured bronzes be at least gated behind MV?

That way people are still forced to use TFC's crucible and Mixer to create the coloured bronzes in the Steam-LV age, but once the player is progressed far enough into the game they could then easily make the coloured bronze ingots with an MV alloy smelter or Multi Smelter in HV.


Just a request so that the late-game player has a way to make these alloys other than the two-step processes of mixer-furnace or ABF-solidifier.

PR #3585, opened 2026-03-28
Shader mod changes

woopSmile:
## What is the new behavior?
added complementary to respect our agreement with EminGT 
Euphoria patches
Supplemental patches

removed colorwheel patcher because our shader natively support colorwheel

## Implementation Details
I think i did it correctly with pakku but id be more comfortable if someone double checked

### DIFF
--- a/pakku.json
+++ b/pakku.json
@@ -256,9 +256,6 @@
         "colorwheel": {
             "side": "CLIENT"
         },
-        "colorwheel-patcher": {
-            "side": "CLIENT"
-        },
         "autopack-director": {
             "export": false
         },
@@ -303,9 +300,19 @@
         "ftb-xmod-compat": {
             "export": true
         },
-
-
         "probejs": {
+            "side": "CLIENT",
+            "export": false
+        },
+        "supplemental-patches": {
+            "side": "CLIENT",
+            "export": false
+        },
+        "euphoria-patches": {
+            "side": "CLIENT",
+            "export": false
+        },
+        "complementary-unbound": {
             "side": "CLIENT",
             "export": false
         }
--- a/pakku-lock.json
+++ b/pakku-lock.json
@@ -2876,67 +2876,6 @@
             ]
         },
         {
-            "pakku_id": "91kPFa77FhKsGmSh",
-            "type": "MOD",
-            "side": "CLIENT",
-            "slug": {
-                "curseforge": "colorwheel-patcher",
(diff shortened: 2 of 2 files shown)


Xikaro:
We need to release our shader version first.  I also don't understand the point of these mods, do they allow us to use the functionality that we transfer and implement in our shaders?

woopSmile:
it is to satisfy the agreement we have with those developers. we get to share our shader only if we install their mods

pakku.json
```diff
@@ -303,11 +300,21 @@
...
+        },
+        "supplemental-patches": {
+            "side": "CLIENT",
+            "export": false
+        },
+        "euphoria-patches": {
+            "side": "CLIENT",
+            "export": false
+        },
+        "complementary-unbound": {
```
Xikaro, line 315: Redefining the "side" is not required here, shaders are already on the client.

And it makes sense to disable exports if they don't get anywhere at all. Then you don't have to add them at all.

Pyritie (approved):

Pyritie:
it's better to be on good terms with the shader authors so we can get help from them in the future

pakku.json
Xikaro: it has not changed

Pyritie: what should be changed? isn't it good to keep `export:false`?

Xikaro: If export:If it's false, they won't get into the modpack for the players, but as I understand it, it's the opposite.

Pyritie: My understanding is that these mods aren't actually required for our shaders to work, but including them like this will still count as "part of the modpack" for crediting/paying the mod authors

Xikaro: Well, with this configuration, they will only be in the development environment. Players will not receive any downloads and, accordingly, the authors will not receive credit.

Xikaro: https://juraj-hrivnak.github.io/Pakku/config-file.html#project_export

Pyritie: With `export:false`? I think this marks the mod as optional, but curseforge does not understand optional, so it always downloads it anyway. Only modrinth/prism understand optional.
But ok, we can remove them from the pakku file so they are always downloaded.

Xikaro: If we follow this logic, then probejs should be there, but it is not
Screenshot_108

Pyritie: ok ok

Pyritie: fixed

PR #3607, opened 2026-03-29
add pt_br lang to ad_astra

gifpxto:
I adapted the PR from Ad Astra by @PrincessStelllar for compatibility to the modpack. Thanks @PrincessStelllar!!


Pyritie:
we do translations via crowdin now, so this will just be overwritten if we accept it

PR #3615, opened 2026-03-29
Add Create millstone recipes for TFC grain, to match quern

lickorice:
## What is the new behavior?
Enables milling grains to flour using Create's millstone. Just to have a bit of quern parity (and more usefulness to the millstone during the early game, other than grinding flux/ore/other dusts).

## Implementation Details
Simply piggybacks a `recipes.create.milling` event in the `recipes.food.js` for-loop for each `TFC_GRAINS`. I put it next to the grain to flour comment in the same file so it's easy to reference later on.

## Outcome
Enables grain to flour recipes for Create's millstone.

### Discord details
lickorice
319285994253975553

### DIFF
--- a/kubejs/server_scripts/tfg/food/recipes.food.js
+++ b/kubejs/server_scripts/tfg/food/recipes.food.js
@@ -145,6 +145,10 @@ function registerTFGFoodRecipes(event) {
 			itemOutputs: [`2x tfc:food/${grain}_flour`],
 			itemOutputProvider: TFC.isp.of(`2x tfc:food/${grain}_flour`).copyOldestFood()
 		})
+    event.recipes.create.milling(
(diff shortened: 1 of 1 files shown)


Pyritie:
We intentionally don't have these because create doesn't understand TFC's food system, and will do things like let you turn fresh grains into rotten flour and rotten grains into fresh flour. Unfortunately it's very complicated to fix this without basically writing an entire millstone from scratch.

Pyritie:
for reference:
- Quern is a TFC thing so it understands this
- Crafting table recipes have TFC-provided methods to handle this (which is where we have mortar and pestle recipes)
- Food Processor recipes are a custom GT machine we created to specifically handle this (same with the food oven, and it's why all food-related recipes are in just those two machines instead of the macerator/mixer/etc)

Issue #3619, opened 2026-03-30
[🐛]: Copper Shingles recipe doesn't exist


Commander-Cat101:
### Was this bug reported before?

- [x] No, it was not.

### Are you using Prism or Curseforge?

No, I am not using Prism or Curseforge.

### Have you tried redownloading the pack?

- [x] Not Applicable / Yes, I have tried downloading a new instance.

### Computer OS

Windows 11

### Modpack Version Used

0.11.28

### Minecraft Environment

Singleplayer

### New Worlds

Yes

### Please provide a link to your latest.log, if relevant.

Not Relevant

### Expected Behavior

The copper shingles family of blocks are craft-able and have recipes show up in REI

### Actual Behavior

Copper shingles family of blocks don't have any crafting recipes and are unobtainable :(

### Steps to Reproduce

1) Load up singleplayer world
2) Notice that copper shingles are uncraftable

nixiee0:
If you don't mind, I'm ready to implement it @Pyritie 

I'm guessing he talking about this block
<img width="430" height="171" alt="Image" src="https://github.com/user-attachments/assets/868546a8-9794-4f61-98a7-deb6697053a9" />

Pyritie:
sure, I think stonecutter recipes should be ok

PR #3630, opened 2026-03-30
Added recipes for copper shingles, tiles and their derivatives

nixieeq:
# I implemented all the same functions (except for recycling) for copper shingles and tiles as for copper cut
>  issue: https://github.com/TerraFirmaGreg-Team/Modpack-Modern/issues/3619

There might be some wrong code somewhere, so it would be nice if you could check it.
>I don't really like messing with legacy code, but it seems like it would be necessary here.
Maybe in the future I would rewrite it completely, because I don't really like it

image
image
image
image


kubejs/server_scripts/minecraft/recipes.js
```diff
@@ -1,91 +1,147 @@
...
+ *
+ * @param {Internal.RecipesEventJS} event
+ * @param {Object} source
+ * @param {Object} target
+ */
+const addCopperWaxingRecipes = (event, source, target) => {
+	for (const [componentName, recipeSuffix] of MINECRAFT_COPPER_RECIPE_COMPONENT_FIELDS) {
+		if (!source[componentName] || !target[componentName]) continue
+
+		event.recipes.gtceu.assembler(`tfg:minecraft/waxing_${recipeSuffix}_${target.name}`)
```
Pyritie, line 61: do these need to be assembler recipes? there's no crafting table recipe they're replacing, so just having them as crafting table recipes or something would be fine imo

kubejs/server_scripts/minecraft/recipes.js
```diff
@@ -1,91 +1,147 @@
...
+const addCopperOxidizingRecipes = (event, source, target) => {
+	for (const [componentName, recipeSuffix] of MINECRAFT_COPPER_RECIPE_COMPONENT_FIELDS) {
+		if (!source[componentName] || !target[componentName]) continue
+
+		event.recipes.gtceu.chemical_reactor(`tfg:minecraft/oxidizing_${recipeSuffix}_${source.name}`)
+			.itemInputs(source[componentName])
+			.inputFluids("#tfc:any_water 150")
+			.circuit(1)
+			.itemOutputs(target[componentName])
+			.duration(1000)
```
Pyritie: the duration can probably be lowered

kubejs/server_scripts/minecraft/recipes.js
```diff
@@ -1,91 +1,147 @@
...
+/**
+ *
+ * @param {Internal.RecipesEventJS} event
+ * @param {Object} element
+ */
+const addCopperVariantAssemblerRecipes = (event, element) => {
+	for (const [sourceName, targetName, recipeSuffix] of MINECRAFT_COPPER_VARIANT_ASSEMBLER_FIELDS) {
+		if (!element[sourceName] || !element[targetName]) continue
+
+		event.recipes.gtceu.assembler(`tfg:minecraft/assembler/${recipeSuffix}_${element.name}`)
```
Pyritie: why assembler? these can just be stonecutter, like how I did all the rock-related stonecutter recipes. For them, people can freely swap between a block, wall, and stair, and then turn any of them into two slabs.

nixiee0:
Look, I've removed the cut copper recipe from the workbench. Now you can make 3 types from ordinary copper block: shingles, tiles and cutted. Already from shingles and so on. you can make ladders and slabs, both on a stonecutter and in an assembler. As for the duration. I didn't change it, it was the Exzeption code, as I understand it, if you want, I can reduce the time. @Pyritie

nixiee0:
In assembler you can make from 4 blocks = 3 stairs, but there is also an option to do 1 to 1 in stonecutting. This is like an incentive for using assembler, this is also Exception code

nixiee0:
It just seemed to me that it wouldn't be logical that there are such recipes for copper cut blocks. and for the rest they will only be in stone cutting

Pyritie:
> As for the duration. I didn't change it, it was the Exzeption code, as I understand it, if you want, I can reduce the time. @Pyritie

please do!

> In assembler you can make from 4 blocks = 3 stairs, but there is also an option to do 1 to 1 in stonecutting. This is like an incentive for using assembler, this is also Exception code

Don't worry about old Exception code. I think consistency with other decoration blocks is better. What do you think?

Pyritie (dismissed):

Pyritie (approved):

Redeix (approved):

PR #3661, opened 2026-04-01
Honeyed apples from Create

Froffy025:
## What is the new behavior?
Adds a crafting/food processor recipe for combining apples & raw honey for a preserved, higher-satiety apple using the Create item that is already in the game.

## Implementation Details
Removed create:honeyed_apple from disabled items list, added food details to data.food.js and tfc:foods/fruits tag.

## Outcome
New food item. The additional nutrition is similar to what is added in porridge with honey. Quick and simple to prepare, and rewards a player with a well-established base

## Additional Information
image
image

For comparison, apple jam is more efficient for bulk crafting, stores better, and can be used in sandwiches... but requires jars, which are a pain in the ass to make before industrialization.

None that I know of...

Froffy

### DIFF
--- a/kubejs/server_scripts/create/recipes.js
+++ b/kubejs/server_scripts/create/recipes.js
@@ -2483,13 +2483,4 @@ const registerCreateRecipes = (event) => {
 	event.smelting('#forge:ingots/tin', 'create:crushed_raw_tin')
 	event.smelting('#forge:ingots/lead', 'create:crushed_raw_lead')
 
-	// apple
-
-	event.shapeless('create:honeyed_apple', ['#tfc:foods/apples', 'firmalife:raw_honey'])
-	
-	event.recipes.gtceu.food_processor('create:honeyed_apple')
-	.itemInputs('#tfc:foods/apples', 'firmalife:raw_honey')
-	.itemOutputs('create:honeyed_apple')
-	.duration(60)
-	.EUt(16)
 }
--- a/kubejs/server_scripts/create/recipes.js
+++ b/kubejs/server_scripts/create/recipes.js
@@ -2486,4 +2486,10 @@ const registerCreateRecipes = (event) => {
 	// apple
 
 	event.shapeless('create:honeyed_apple', ['#tfc:foods/apples', 'firmalife:raw_honey'])
+	
+	event.recipes.gtceu.food_processor('create:honeyed_apple')
+	.itemInputs('#tfc:foods/apples', 'firmalife:raw_honey')
+	.itemOutputs('create:honeyed_apple')
+	.duration(60)
+	.EUt(16)
 }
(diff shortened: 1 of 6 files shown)


kubejs/server_scripts/create/tags.js
```diff
@@ -112,6 +112,8 @@ const registerCreateItemTags = (event) => {
 	event.add('forge:smooth_stone_slab', 'create:polished_cut_scoria_slab')
 	event.add('forge:smooth_stone_slab', 'create:polished_cut_scorchia_slab')
 	event.add('forge:smooth_stone_slab', 'create:polished_cut_veridium_slab')
+
+	event.add('tfc:foods/fruits', 'create:honeyed_apple')
```
Redeix, line 117: Also needs the `tfc:foods` tag

kubejs/server_scripts/tfg/food/data.food.js
```diff
@@ -596,6 +596,15 @@ function registerTFGFoodData(event) {
 		food.saturation(1);
 	});
 
+	// Honeyed Apple
+	event.foodItem('create:honeyed_apple', (food) => {
+		food.hunger(3.5);
+		food.decayModifier(0.8);
```
Redeix, line 602: This decay modifier is a bit extreme for a resource as cheap as honey. Halving the current apple decay rate should be sufficient.

kubejs/server_scripts/create/recipes.js
```diff
@@ -2482,4 +2482,14 @@ const registerCreateRecipes = (event) => {
 	event.smelting('#forge:ingots/silver', 'create:crushed_raw_silver')
 	event.smelting('#forge:ingots/tin', 'create:crushed_raw_tin')
 	event.smelting('#forge:ingots/lead', 'create:crushed_raw_lead')
+
+	// apple
+
+	event.shapeless('create:honeyed_apple', ['#tfc:foods/apples', 'firmalife:raw_honey'])
+	
+	event.recipes.gtceu.food_processor('create:honeyed_apple')
+	.itemInputs('#tfc:foods/apples', 'firmalife:raw_honey')
```
Redeix: This recipe has no item stack modifier information. It can be moved to the TFG food recipes file so you can see examples of how to use ISPs. Also I don't believe that this will not cause conflicts without a circuit number. There are exiting foods that use an apple and a sweetner

Redeix (changes requested):

kubejs/server_scripts/tfg/food/data.food.js
Redeix: What I mean is double how long apples last. So from like 8 days to 16 or whatever

Froffy025:
thankyouthankyou o7

kubejs/server_scripts/tfg/food/data.food.js
Froffy025: image
with no other modifiers, red apples currently last 13(.5?) days, and honeyed apples currently last 27 days, so it is roughly doubling the decay time here! green apples weirdly last less time, at only 8 days, but i think this might just be an oversight? is it okay to stick with the red apple as a baseline here?

Redeix: Oh wow I didn't realize red apples lasted so long. Usually fruit expire quickly. In that case I think it's fine.

Pyritie (approved):

Redeix (approved):

PR #3708, opened 2026-04-05
Switch from default options to kubejs

### LINKED ISSUE #3660: [❓]: KubeJs Makes Default Options Mod Obsolete
### Was this question asked before?

- [x] No, it was not.

### Your Question

Quoting [KubeJS](https://kubejs.com/wiki/tutorials/default-options) directly:

> You can ship default options from options.txt with KubeJS. This includes keybindings, video settings, enabled resource packs, controls like autojump and toggle sprint and weird things like advanced tooltips.
Why use this instead of just shipping options.txt? If you ship options.txt then the users options will get overridden every time they update your modpack, where-as KubeJS only sets the options once, on the first time the modpack boots.
To use it simply make a file called defaultoptions.txt in the kubejs/config folder. Then copy any lines you want to set by default over from the normal options.txt file. You can also just copy the entire file if you want to include everything.
A full list of what options the options.txt file can contain is available on the [Minecraft Wiki](https://minecraft.wiki/w/Options.txt).

Redeix:
I just copied the strings from default options to kubejs. Some of the keybinds confuse me idk.
resolves #3660


SakuraKitsurugi:
But, doesn’t default options mod ship mod options?

Pyritie:
> But, doesn’t default options mod ship mod options?

no, we ship the config files for other mods

Pyritie (approved):

PR #3747, opened 2026-04-09
CTM for Ad Astra Blocks

### LINKED ISSUE #3581: [💎]: Ad astra connected texture blocks

- [x] No, it was not.

Finish the connected textures for those ad astra blocks @SakuraKitsurugi 

### Why is this needed?

as a reminder :p

SakuraKitsurugi:
### What is the new behavior?
Adds connected textures to certain Ad Astra blocks.

### Implementation Details
- Created PNGs and MCMETAs in `assets/ad_astra/textures/block/`

### Additional Information
- Due to the way LDLIB work - I can't add connected textures to directional blocks.
- Closes #3581


Pyritie (approved):

Redeix:
I would have wished that if we were going to do this then we would have replaced their textures with our own unique but similar ones. As it is we have to replace all their assets in order to fork the project. I don't want to deal with their license restrictions further

Pyritie:
were we still going to actually fork it? I thought the plan was to keep AA but just mixin out the parts of it we don't want

Redeix:
Whatever we end up doing I'd rather not continue to make new things that depend on the textures at the very least. It's just making it harder and harder if in the future we decide to switch away

PR #3770, opened 2026-04-12
For correct translation | gtceu

nixiee0:
To ensure a holistic translation, we need to add all tiers for all tools.

It seems like it should work, but just in case, check if it can be implemented this way.


Pyritie:
could you put this over here as well please https://github.com/TerraFirmaGreg-Team/Tools-Modern/tree/dev/LanguageMerger/LanguageFiles/gtceu/en_us

Pyritie (approved):

PR #3835, opened 2026-04-18
Fix lutetium position

ashleney:
image


Pyritie:
don't you need to move the square too?

ashleney:
I didn't expect the PR to open right away when editing a file from the browser, sorry

Pyritie (approved):

PR #3850, opened 2026-04-19
popcorn

applenper:
popcorn and buttered popcorn food item
image
image


Pyritie (dismissed):

Redeix (approved):
We can add this. But let's not accept anymore food items while I'm in the middle of a massive food code base refactor please 🥺

PR #3861, opened 2026-04-21
Cherry Blossom Leaves

SakuraKitsurugi:
### What is the new behavior?
Adds Minecraft Cherry Leaves to be used as a decorative block.

### Implementation Details
- Added new texture to be similar to TFC leaves.
- Modified the leaves to imitate TFC leaves by giving them no collision and reducing movement speed.
- Modified leaves' loot to drop itself and only when destroyed with `#forge:shears`
- Obtainable only with the EGH by using cherry saplings. Does not give anything but the leaves.

### Additional Information
- Texture done by @Redeix
- Will modify the name of the leaves to "Cherry Blossom Leaves" in Tools repo to distinguish from TFC's Cherry Leaves.


Pyritie (commented):
you'll probably also want to remove them from the startup script for minecraft constants, so the leaves are visible in jei

SakuraKitsurugi:
I did? At least I think

SakuraKitsurugi:
[kubejs/startup_scripts/minecraft/constants.js](https://github.com/TerraFirmaGreg-Team/Modpack-Modern/pull/3861/changes#diff-6c2bce5ef16a68a7e6fe7d871bfa156dca7cf4fdd57b95f66ef53ca9264c5180)

Redeix (approved):

Pyritie (approved):
somehow I completely missed that

PR #3896, opened 2026-04-26
fix(food): Make brioche dough craft output 5

pehala:
## Summary
Recipe had 6 before, but in practice it outputted 5 (verified on local server), as mixing bowl has maximum capacity of 5 (AFAIK). Processor recipe left unchanged
I reduced the recipe to output 5 so it matches the real world. I believe it is harmless.
I did not test this change, but given it is simple kubejs update, I dont think I need to

TheSprtCZ


Redeix:
This is already fixed in the food rework branch. Thank you though

PR #3906, opened 2026-04-28
Create Fluid Logistics Migration

### LINKED ISSUE #3727: Migrate to  Create Fluid Logistics to replace Create Factory Logistics
> Now that create:factory logistics is not longer supported is it worth looking at using a different mod for create fluid logistics?
> I’ve been testing with this one https://modrinth.com/mod/createfluidlogistic, its not quite ready to be used in the pack (there are some small issues) but i’ve found it much more intuitive than create: factory logistics. The main advantage is that packages that contain items and fluids actually work.
> 
> If there is interest I can do more testing and make the recipes integrate with the rest of the pack
> 
> Obviously if we change to this mod it would break everyone setups so if you want to, do the change during a major version 

 _Originally posted by @Bumperdo09 in [#3700](https://github.com/TerraFirmaGreg-Team/Modpack-Modern/issues/3700#issuecomment-4193405555)_

Bumperdo09:
## What is the new behavior?
Migrate to Create Fluid Logistics
For https://github.com/TerraFirmaGreg-Team/Modpack-Modern/issues/3727

Also changes to Core-Modern to go along with this

## Outcome

New fluid packager recipe. Item on the right is waterproof cardboard, crafted by 1 cardboard and 1 wax
image

New block: Multi-Fluid Tank. can hold up to 8 different types of fluids at a time, but a total volume max of 8 buckets (per tank block)
image

New block: Multi-Fluid Access Port. Its like the Inventory Access Port but for fluids
image

New Item: Hand Pointer. It allows various configuration options for create logistics components like force-powering packagers and re-targeting frog ports after they have been placed
image

- Open to suggestions for changing recipes of new items
- Changed some of the create logistics quests to make sense given the the new mod
- Added assembler recipes for the create packager and the new fluid packager because they where the only create logistics item that didn't have one. They are both on circuit 20, the same as the frog port.
- The portable fuel interface recipe is changed as the fluid mechanism was apart of Create Factory Logistics, and so is now gone

## Additional Information
This shouldn't be merged until the next alpha release to limit the breaking of peoples worlds

Pakku needs to be updated too (removing "create_factory_logistics", adding "createfluidlogistic")


kubejs/server_scripts/create/recipes.js
```diff
@@ -1784,6 +1784,14 @@ const registerCreateRecipes = (event) => {
 		E: 'create:electron_tube'
 	}).addMaterialInfo().id('tfg:create/shaped/packager')
 
+		event.recipes.gtceu.assembler('create:packager')
+		.itemInputs('#forge:small_gears/red_alloy', '#forge:springs/wrought_iron', 'create:copper_casing', 'create:bound_cardboard_block', 'create:electron_tube')
+		.itemOutputs('create:packager')
+		.circuit(20)
+		.duration(100)
+		.EUt(GTValues.VA[GTValues.LV])
+		.addMaterialInfo(true)
```
AirRice:
```suggestion
```
You shouldn't add assembler recipes if the crafting recipe doesn't need GT tools as they are the main reason for them existing

kubejs/server_scripts/create_fluidlogistics/recipies.js
```diff
@@ -0,0 +1,93 @@
+// priority: 0
+"use strict";
+
+function registerCreateFluidLogisticsRecipes(event) {
+
+	event.remove({ mod: "fluidlogistics" });
```
AirRice:
```suggestion
    event.remove({
		not: [
			{ id:'fluidlogistics:horizontal_multi_fluid_tank_from_tank' },
			{ id: 'fluidlogistics:multi_fluid_tank_from_horizontal' },
			{ id: 'fluidlogistics:hand_pointer_clear' },
		], mod: "fluidlogistics" })
```
Probably better to just carry over these recipes from the mod than delete & re-define them later

kubejs/server_scripts/create_fluidlogistics/recipes.js
```diff
@@ -0,0 +1,93 @@
...
+"use strict";
+
+function registerCreateFluidLogisticsRecipes(event) {
+
+	event.remove({ mod: "fluidlogistics" });
+
+    event.shapeless('fluidlogistics:waterproof_cardboard_block', ['create:cardboard_block', '#forge:wax'])
+    event.recipes.create.item_application('fluidlogistics:waterproof_cardboard_block', ['create:cardboard_block', '#forge:wax'])
+    
+	event.recipes.gtceu.packer('tfg:waterproof_cardboard_block')
```
AirRice, line 16: This being a packer recipe seems odd to me since the packer isn't often used for this kind of operation. Usually item application recipes are 1:1 transferred to assembler recipes

kubejs/server_scripts/create_fluidlogistics/recipies.js
```diff
@@ -0,0 +1,93 @@
...
+	], {
+		A: '#forge:bolts/brass',
+		B: '#forge:tools/screwdrivers',
+		C: '#forge:plates/brass',
+		D: '#forge:glass_panes',
+		E: '#forge:tools/wrenches'
+	}).id("fluidlogistics:shaped/multi_fluid_tank")
+
+	event.recipes.shapeless('fluidlogistics:horizontal_multi_fluid_tank', 'fluidlogistics:multi_fluid_tank').id("fluidlogistics:shapeless/multi_fluid_tank_conversion_from_normal")
+	event.recipes.shapeless('fluidlogistics:multi_fluid_tank', 'fluidlogistics:horizontal_multi_fluid_tank').id("fluidlogistics:shapeless/multi_fluid_tank_conversion_from_horizontal")
```
AirRice:
```suggestion
```
Don't need to add these if you just keep the mod's recipes for them

kubejs/server_scripts/create_fluidlogistics/recipies.js
```diff
@@ -0,0 +1,93 @@
...
+
+	event.recipes.gtceu.assembler('fluidlogistics:fluid_packager')
+		.itemInputs('#forge:small_gears/red_alloy', '#forge:springs/wrought_iron', 'create:copper_casing', 'fluidlogistics:waterproof_cardboard_block', 'create:electron_tube')
+		.itemOutputs('fluidlogistics:fluid_packager')
+		.circuit(20)
+		.duration(100)
+		.EUt(GTValues.VA[GTValues.LV])
+		.addMaterialInfo(true)
```
AirRice:
```suggestion
```

As mentioned above, shouldn't add an assembler recipe if a crafting recipe exists that isn't GT tool-limited

config/ftbquests/quests/chapters/create_logistics.snbt
```diff
@@ -284,13 +285,13 @@
 				}
 				{
 					id: "38DC47D44D9741FC"
-					item: "create_factory_logistics:factory_fluid_gauge"
+					item: "create:factory_gauge"
 					optional_task: true
 					type: "item"
 				}
```
AirRice:
```suggestion
```

You can just remove this since it was for the quest to recognise both fluid and factory gauges, doesn't make sense for it to be duplicated anymore

AirRice (commented):
All seems good to me at a glance with comments left on parts I think should be changed.

Have we considered if we want the mod's newly added portable stock ticker? I haven't played around with it but it seems ok as long as it doesn't remotely allow people to get packages and still requires setting up package routing to receive something to an address. If so we should add its recipe (crafting together a stock ticker with a GT emitter/sensor, LV would be how I add it) and un-remove its clearing recipe.

kubejs/server_scripts/create/recipes.js
Bumperdo09: I talked with Pyrite about this, packagers are the only create logistics component needed in large quantities that dont have a GT assembler recipe. In my play through packagers and wood slabs where the only two recipes that made me setup a 3x3 mechanical crafting setup

kubejs/server_scripts/create_fluidlogistics/recipes.js
Bumperdo09: I was following a similar recipe for bound cardboard (4x cardboard + 1x string -> 1x bound cardboard)

kubejs/server_scripts/create/recipes.js
TomPlop: If it has no tool universally I would prefer not having an assembler recipe. If you want to automate you can make mechanical crafter or just batch craft them that's totally doable with Create logistics.

AirRice: Given most players are going to be slowly going from no auto crafting -> create logistics auto crafting with gauges -> AE2, I think it's good to force them to automate 3x3 crafting, since it'll lead them into setting up molecular assemblers down the line. With enough rpm shaped 3x3 crafters are faster than almost any assembler recipe anyway

Bumperdo09:
The portable stock ticker from fluidlogistics isnt actually in the mod for the forge version yet. I did some testing with it, it just opens the normal stock ticker gui, you still input the address to send the packages to and have to build the package routing like normal, it doesnt teleport them to you. The main annoyance I had with it was that there was no way of setting up categories

If it eventually gets added to the forge version of the mod I think we should add it, it seems atleast some what usefull

technoch1ef:
@Bumperdo09, regarding stock tickers. I checked CC peripherals and the old Fluid Logistics always returned fluids as air, which was super annoying. Any way we can make it work with [StockLinkPeripheral](https://github.com/Creators-of-Create/Create/blob/mc1.20.1/dev/src/main/java/com/simibubi/create/compat/computercraft/implementation/peripherals/StockTickerPeripheral.java)

I can also chip in if help needed, but I'm not that good with Java

Example
```log
--- Entry #1 ---
  name = tfc:sand/black
  itemGroups = {
    (empty)
  }
  tags = {
    forge:sand = true
    tfc:volcanic_sand = true
    minecraft:sand = true
  }
  count = 80
  maxCount = 32
  displayName = Black Sand

--- Entry #2 ---
  name = tfc:mortar
  itemGroups = {
    (empty)
  }
  tags = {
    tfc:mortar = true
  }
  count = 576
  maxCount = 64
  displayName = Mortar

--- Entry #3 ---
  tags = {
    (empty)
  }
  name = minecraft:air
  itemGroups = {
    (empty)
  }
  maxCount = 64
  count = 60000
  nbt = 41883520c3071f5f4a4a4613fb005e0c
  displayName = Air

--- Entry #4 ---
  tags = {
    (empty)
  }
  name = minecraft:air
  itemGroups = {
    (empty)
  }
  maxCount = 64
  count = 16000
  nbt = 41883520c3071f5f4a4a4613fb005e0c
  displayName = Air
```

Bumperdo09:
> @Bumperdo09, regarding stock tickers. I checked CC peripherals and the old Fluid Logistics always returned fluids as air, which was super annoying. Any way we can make it work with [StockLinkPeripheral](https://github.com/Creators-of-Create/Create/blob/mc1.20.1/dev/src/main/java/com/simibubi/create/compat/computercraft/implementation/peripherals/StockTickerPeripheral.java)
> 
> I can also chip in if help needed, but I'm not that good with Java

I am a bit confused.

Does the new mod (fluid logistics) work properly with CC? the old mod (Factory Logistics) doesn’t really matter for us anymore.

If fluidlogistics is not working with CC you should help add the compatibility directly to https://github.com/yision1/CreateFluidLogistic

As for helping with this PR, it’s pretty much done. I am just waiting on the next alpha release of the pack

Bumperdo09:
What are peoples thoughts on needing tools to craft packagers?
image
image

I think would make more sense for them to require tools because pretty much all other create logistics items require tools. It also means there can be an assembler recipe for easier auto crafting of them.

kubejs/server_scripts/create_fluidlogistics/recipes.js
```diff
@@ -0,0 +1,90 @@
+// priority: 0
```
SakuraKitsurugi, line 1: Not a huge issue, but this file's name has a typo

SakuraKitsurugi (dismissed):

SakuraKitsurugi:
you will need to resolve merge conflicts too

.gitignore
Pyritie, line 1: what is `dump/`?

Bumperdo09: It was from when I was trying to dump recipes to automatically check for assembler recipes. I accidentally committed it (I didn’t know about the local git-ignore). I can remove it

SakuraKitsurugi (approved):

Pyritie (approved):

PR #3914, opened 2026-04-30
medicine powders, cheaper salves

applenper:
4 new early game medicines in the form of powder with quirky effects, crafted with pitch and bonemeal

salves are now crafted with creosote rather than phenol, along with mixing bowl recipes, allowing salves to be created in iron age

image
^that's slowness IV

image

image

^(i originally had acrobat powder give slowfalling hence why it accepts slowfalling ingredients rather than speed, i decided to keep it that way since it gives more plants more uses)

image


Pyritie (approved):

TomPlop (approved):
All good for me but I wonder if we could add hot spring for one or two? maybe the weakest one, I just think it would be nice to make it more useful early on

Redeix (approved):

Issue #3913, opened 2026-04-29
[💎]: Processing Kaolin Clay


deadlyjackattack:
### Was this enhancement suggested before?

- [x] No, it was not.

### What would you like to be added/ changed?

One point of friction regarding kaolin is processing. The best method available using pre-steel technology is putting Kaolin Clay balls into small Ceramic Vessels, and baking these in pit kilns.

I like kaolin and what it does for early TFG gameplay. But I think this method is obtuse for a newer player. They will likely do what I have seen many do: use a Forge to process the first handful, then a Crucible for the rest.

I think the more intuitive method should be the more practical method.

Potential solutions:

- Allow blocks of Kaolin Clay to be baked into Kaolinite directly.

- Add a recipe for a "Kaolin-Packed Vessel."
  - Heats faster and holds more than a regular vessel containing kaolin clay.
  - Single-use; must be smashed (used? thrown? hammered?) to release its contents.

> The idea behind the packed vessel is that 1. the vessels are often not used again, or simply destroyed anyway, and 2. the recipe would be readily visible in EMI.
> The practical version would be a static handful of recipes using blocks of kaolin (or a compressed intermediate) to make the vessels.
> The elegant (but maybe technically unfeasible or unworthwhile) version would allow a vessel to accept balls or blocks of kaolin in variable stack sizes, keeping track of its contents.

### Why is this needed?

The goal is just to make things a bit smoother. There is already much trouble to obtain the clay from cold and dry places. Adding one or more alternative recipes that are readily visible in EMI would guide people to more pleasant options. I know someone who used firepits to bake their kaolin and they regret it immensely.

Pyritie:
I'm bad at math, how much kao powder should a block turn into? 80% of 1? 20% of 4? I want it to be the same ratio as the single clays

SokvisalMong:
One way is to treat them as their own chances. Since 1 Clay Block equals 4 clay balls, the math would be **4 × 20% = 80%** which gives you an 80% chance of obtaining 1 powder. 

This also means 16 Clay Blocks would be **16 × 4 × 20% = 1280%**. Guaranteeing 12 powder with an 80% chance to get an extra.

The other way is to treat them all as individual clay balls and use a range table, but I feel like that's too much math and probably prefer the first method.

Pyritie:
I don't think the range is even possible to do in the heating recipe lol

SokvisalMong:
[...]

PR #3922, opened 2026-04-30
added kaolin clay blocks to be processed into kaolinite powder

SokvisalMong:
## What is the new behavior?
Able to melt kaolin blocks for powder

## Implementation Details
Added additional item tags for kaolin clay blocks and recipes for alloy smelter and heating at an increased chance.

## Outcome
#3913 

## Additional Information
Should probably nerf the numbers a bit for heating process since I put the same heating point as kaolin clay (500). Maybe increasing it higher so it takes additional time to melt in crucibles and such.

image

image

image


kubejs/server_scripts/tfc/tags.js
```diff
@@ -20,9 +20,6 @@ function registerTFCItemTags(event) {
         }
     });
 
-    // Add chainsaws to #tfc:axes so they drop fruit tree saplings
```
Pyritie, line 23: can you add this back?

Pyritie (changes requested):
do the kaolin blocks already have a heat definition? that's the part of tfc that says how fast stuff heats up, and is required to use them in any heating recipes. If you need to add it, you can do it in server_scripts/tfg/data.js I think?

kubejs/server_scripts/tfc/tags.js
SokvisalMong: oh oops ok

SokvisalMong:
I forgot which one but since tfc uses either Weight or Size for how fast stuff heats up, then they're already different and should have different time to heat up.

So maybe no need for additional changes?

Pyritie:
no, Weight determines stack size and Size determines which containers can hold it. Neither is related to heat data. Have you checked that the recipes function ingame? It wouldn't be the first time we've added heating recipes and forgot the heat data lol

SokvisalMong:
Yeah I'm just checking it now and realized I can't even place it in the container to heat it up lol. Adding it now

SokvisalMong:
Checking the source of tfc, the original heat capacity for kaolin clay is 2.0, is 3.0 for blocks ok?

Pyritie:
whatever feels right

SokvisalMong:
I've tested values between 2.5 and 4.0 and 3.0 seems like an ok middle point

Pyritie (approved):

Redeix (approved):

PR #3953, opened 2026-05-03
Crafting Station Whitelist

### LINKED ISSUE #3946: [💎]: Remove Crafting Station from the modpack

- [x] No, it was not.

The Crafting Station should be removed from the modpack as soon as possible

### Why is this needed?

It currently causes several serious issues:

- It can duplicate items. I accidentally duplicated items twice while using it.
- It appears to be one of the main causes of TPS drops on servers.
- It sometimes crashes the game.
- When players connect it to Item Vaults and open it, the server TPS can drop heavily

It also breaks the intended design of the Create storage system. Instead of building proper logistics, item routing, vault access, and automation, players can simply attach a Crafting Station to Item Vaults and bypass most of that gameplay.

Tonight on the official server, we tried to investigate the cause of the TPS drops, and the Crafting Station turned out to be the main issue. Every time it was opened while connected to Item Vaults, the server TPS instantly dropped to around 6

This happened at night with low player activity. During the day, when many more players are online and this block is used by many bases, the impact is probably much worse

SakuraKitsurugi:
### What is the new behavior?
Removes the crafting station blacklist and adds a whitelist instead.

Current Whitelist:
| Block Entity |
| :--- |
|TFC Chest|
|TFC Trapped Chest|
|FramedBlocks Framed Chest|
|FramedBlocks Framed Secret Storage|
|GregTech Crate|

### Additional Information
- Pakku does not take my CF key so please updated Crafting Stations Improved to 1.4.0
- Closes #3946 
- Will stay as a Draft until/if my PR (https://github.com/Deepacat/Crafting-station-improved/pull/7) to Crafting Station's github is accepted and a new update is released.

### DIFF
--- a/defaultconfigs/craftingstation-server.toml
+++ b/defaultconfigs/craftingstation-server.toml
@@ -2,4 +2,6 @@
 [general]
 	#Are side inventories displayed in the crafting grid?
 	"display side inventories in crafting grid" = true
+	#Should Crafting Station only detect whitelisted blocks?
+	"use whitelist for detecting storage blocks" = true
 
--- a/kubejs/server_scripts/crafting_station/tags.js
+++ b/kubejs/server_scripts/crafting_station/tags.js
@@ -9,7 +9,7 @@ const registerCraftingStationBlockTags = (event) => {
 }
 
 const registerCraftingStationBlockEntityTags = (event) => {
-    global.CF_DISABLED_STORAGES.forEach(storage => {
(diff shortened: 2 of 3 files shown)


Pyritie (approved):

Pyritie:
do you want to do the pakku as well, or

SakuraKitsurugi:
I added the comment recently, I am still unable to do it :pensive: (pakku does not take my CF key)

PR #3972, opened 2026-05-06
Fix whisk recycling

ashleney:
Fixes whisks costing 4 ingots but recycling into 7 dust.

Recycling whisks can dupe naquadah and tungsten so I believe it's worth fixing.


TomPlop:
Why do we even need these whisks ?

ashleney:
greate mixers

TomPlop:
I thought you couldn't craft them after EV

Pyritie (approved):

Pyritie:
> Why do we even need these whisks ?

they're also used for all of our food processors

TomPlop:
Aaaaah okay that one makes sense

TomPlop (approved):

PR #3973, opened 2026-05-06
changed cutter slab and lumber recipe and unpack lumber from planks

### LINKED ISSUE #3901: [💎]: Add way of automating wood slabs without 3x3 crafting

- [x] No, it was not.

Recipe changes I propose:

1. replace current plank cutter recipe (which is 1x plank -> 4x lumber) with 1x plank -> 2x slab
    This recipe doesn't seem very useful to begin with, if you want to un-craft planks for some reason there is still a 3x3 crafting recipe to do it (1x saw + 1x plank -> 4x lumber). This change will also match the way to make stone slabs in the cutter. Finally a cutter recipe already exists for 1x slab -> 2x lumber so the original plank un-crafting is still automate-able (for those who want it) but its just needs 2 steps

2. Add a packer recipe for 4x lumber -> 1x planks. This adds a GT way of automatically crafting planks. Currently the only way is with create mechanical press or 3x3 crafting

### Why is this needed?

Wood slabs are required for the wood train tracks, which are needed in large amounts.
With the current recipes there is no way to automate wood slabs without 3x3 crafting (with create mechanical crafters)

In my play through wood slabs where the only reason I built a 3x3 mechanical crafter (and I automated most things starting at LV)

SokvisalMong:
## What is the new behavior?
Alternative method to using Mechanical Crafters to automate Slabs

## Implementation Details
Removed GT Cutter Recipe Plank -> 4x Lumber
Added GT Cutter Recipe Plank -> 2x Slab
Added GT Packer Recipe Plank + Circuit 1 -> 4x Lumber

## Outcome
#3901

### DIFF
--- a/kubejs/server_scripts/tfg/natural_blocks/recipes.wood.js
+++ b/kubejs/server_scripts/tfg/natural_blocks/recipes.wood.js
@@ -83,7 +83,12 @@ function registerTFGWoodenRecipes(event) {
 			)
 				.id(`tfg:shapeless/${name}_lumber_from_plank`)
 
-			generateCutterRecipe(event, plank, `4x ${lumber}`, 50, 7, `${name}_lumber_from_plank`)
+			event.recipes.gtceu.packer(`tfg:${name}_lumber_from_plank`)
+				.itemInputs(`${plank}`)
+				.circuit(1)
+				.itemOutputs(`4x ${lumber}`)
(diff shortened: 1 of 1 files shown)


TomPlop:
Why?

Pyritie (approved):

Redeix (approved):

PR #3977, opened 2026-05-06
set flax loot table to proper age

applenper:


Pyritie:
do sunflowers, cucumbers, and canola work correctly already? those are our other two-tall plants

applenper:
> do sunflowers, cucumbers, and canola work correctly already? those are our other two-tall plants

all other plants seem to work fine as-is, flax is just the only crop that grows to age 6

Pyritie (approved):
that would explain it

Redeix (approved):

PR #3983, opened 2026-05-07
Rebalance the AP inventory manager

ashleney:
As tom previously mentioned, the inventory manager causes issues with players abusing them to avoid using the railgun. This makes it be LuV (the same tier as the end automata) which is when TFG could in the future potentially unlock teleportation (LuV/ZPM).


TomPlop:
Sounds a good idea to me at least temporary until we get into LuV/Zpm

TomPlop (dismissed):

SakuraKitsurugi (approved):

TomPlop (approved):

PR #3984, opened 2026-05-07
Fix barrels in railguns

ashleney:
Fixes a funny exploit where people sent fluid inside barrels.


TomPlop:
lol we missed that one. I wonder though if at that point it's not just easier to use the cells rather that the barrel lol

TomPlop (approved):

ashleney:
should buckets also be blacklisted? on one hand I don't think using a create item drain in your space logistics is intended but on the other this would make it impossible to send any fluids. (idk how to blacklist buckets)

TomPlop:
You can and should use the gt cells, I dunno if buckets should be blacklisted but it sounds like a pretty tedious ways to do it opposed to cells

ashleney:
makes sense, buckets can stay then

SakuraKitsurugi (approved):

PR #4012, opened 2026-05-11
Add sophisticated backpacks to crafting station whitelist

winjwinj:
## What is the new behavior?
Adds backpacks to the whitelist for the crafting station


SakuraKitsurugi:
The whitelist has not been merged into Crafting Stations. Also, Sophisticated Backpacks is NOT going into the whitelist due to constantly causing issues.

Oh and you are trying to PR to main, PRs should go to dev.

PR #4024, opened 2026-05-13
Fix sequenced train track recipes and hide in EMI

### LINKED ISSUE #4023: [🐛]: Impossible to craft normal train tracks with steel rods with create sequenced recipe

- [x] No, it was not.

### Computer OS

Windows 10

### Modpack Version Used

0.12.7

### Minecraft Environment

Forge Server

### New Worlds

Yes

no

### Expected Behavior

Crafting 32 tracks using steel rods instead of wrought iron ones

### Actual Behavior

It won't accept steel rods, only wrought iron ones

Image

The Narrow tracks recipe works

### Steps to Reproduce

Try to craft normal tracks using create 

### Additional Information

_No response_

SakuraKitsurugi:
### What is the new behavior?
- Creates new sequenced items to make steel versions of the recipes properly function. 
- Made all inomplete train tracks be hidden from recipe view (clicking or right clicking on them gives no information).
- Fixed a bug with pyroxenite incomplete tracks being used in create:track sequenced recipe.

### Additional Information
- Fixes #4023
- Moved some stuff around in the railways file just in case I need to work with it later.

### DIFF
--- a/kubejs/server_scripts/railways/recipes.js
+++ b/kubejs/server_scripts/railways/recipes.js
@@ -535,6 +535,67 @@ const registerRailWaysRecipes = (event) => {
 		.duration(250)
 		.EUt(32)
 
+	// Create Stone Tracks (Normal)
+	const stoneTrackSlabs = Ingredient.of('#tfg:rock_slabs').subtract('minecraft:blackstone_slab')
+
+	event.recipes.createSequencedAssembly([
+		'32x create:track',
+	], stoneTrackSlabs, [
+		event.recipes.createDeploying('railways:track_incomplete_blackstone', ['railways:track_incomplete_blackstone', stoneTrackSlabs]),
+		event.recipes.createDeploying('railways:track_incomplete_blackstone', ['railways:track_incomplete_blackstone', '#tfg:track_rods']),
+		event.recipes.createDeploying('railways:track_incomplete_blackstone', ['railways:track_incomplete_blackstone', '#tfc:mortar']),
+		event.recipes.greate.pressing('railways:track_incomplete_blackstone', 'railways:track_incomplete_blackstone'),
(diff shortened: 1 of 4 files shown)


Pyritie:
damn is creating all these items really the only way 😭

SakuraKitsurugi:
Sadly, I could try thinking of something else if you give me some time.

Pyritie:
what do you think about something like:
- all rail recipes output 32
- instead of separate recipes for wrought iron and steel rods, put both steel rods and LONG wrought iron rods into a single tag. So this way, steel is still more efficient. Then use this tag in the recipe instead of the individual rods

SakuraKitsurugi:
Not a bad idea, would reduce train track code by a bit too. Will do.

kubejs/server_scripts/railways/recipes.js
Pyritie, line 656: since blackstone slabs are `#tfg:rock_slabs`, how does it know whether to make a blackstone track or normal create track?

SakuraKitsurugi:
i moved create track recipe below everything else which lets blackstone slab be prioritized in making blackstone rails

Pyritie:
I don't think that's how it works. Would it work if you did this [tag subtraction](https://github.com/TerraFirmaGreg-Team/Modpack-Modern/wiki/%5BEN%5D-Dev-Cheat-Sheet) instead?

SakuraKitsurugi:
No, I mean, I tested it out and it works, but sure can change it.

Pyritie:
it's the kind of thing that depends on load order, which isn't very
deterministic

SakuraKitsurugi:
Works ingame, should be correct way of doing it now?

Pyritie (approved):

PR #4031, opened 2026-05-15
Feature/immersive vehicle compatibility

ImmortalEthereal:
## What is the new behavior?
Immersive Vehicle, Official Content Pack and Official Automobile Pack added to the pack.

## Implementation Details
Most items have their recipes, tags and data changed, no quests added yet.

AdAstra might have compatibility issues but probably resolved in the latest release of IV.

ilikecupcakes


Redeix:
I'll check it out soon

Redeix (approved):
I sort of wish the vehicles has a bit more tier diversity instead of all being around steel tier. But besides that its good work for optional compatibility. Theres a few minor things missing like unaccesible items not being hidden. But nothing major.

Redeix:
Updated https://github.com/TerraFirmaGreg-Team/Modpack-Modern/wiki/%5BEN%5D-Suggested-additional-mods

Redeix:
Based on the way tags are set up, each content pack requires each other. But thats okay. I tested removing each mod one at a time and had no issues. The optional compat check needs to look for an official content pack item since it uses the same name space as the base mod.

Pyritie (approved):
interesting mod -- I've not seen one that puts entire recipes in a config file before, or has item IDs with dots in them

PR #4087, opened 2026-05-22
Beneath trees fix

Filiipa:
## What is the new behavior?
_Removed the persistent tag from the leaves._


Pyritie:
there's conflicts

PR #4105, opened 2026-05-23
Update CONTRIBUTING.md

ashleney:


.github/CONTRIBUTING.md
```diff
@@ -1,140 +1,39 @@
...
+## Making changes to TFG
 
-#### Required Software
-- [Git]: A version control system for managing source code.
-- [Pakku]: A tool for dependency management and modpack building.
+### Editing recipes, mod configs and loottables
 
-> [!WARNING]  
-> Ensure that Java version 17 is installed on your machine, as it is required for Forge and Pakku to function properly.
+To edit recipes and configs, you will need to make edits to the `kubejs` directory of your instance. Check out the TerraFirmaGreg's wiki for some guidance on how to use kubejs. To test your changes live in-game, you can simply use the `/kubejs reload ...` command.
```
Pyritie: - configs are not in the kubejs folder
- our wiki is not guidance on how to use kubejs, just the specific few things we've added to it. Linking to the kubejs website is better, even if it's a bit shit

.github/CONTRIBUTING.md
```diff
@@ -273,403 +172,67 @@ There are two approaches to creating a Pull Request: via terminal and via Visual
...
+4. Make a pull request.
 
-> [!TIP]
-> Эта команда загружает все необходимые файлы проекта в папку модпака. 
-> Учтите, что команда может отличаться в зависимости от способа установки Pakku.
+## Setting up a dev instance
 
->[!WARNING]
-> Это не обновит ваш TerraFirmaGreg-Core-Modern! Продолжайте следовать инструкциям.
+For most changes, it is completely optional to set up a dev instance. Refer to the section on making changes to kubejs and to core if you are making a simple change.
```
Pyritie: this section should be the first thing because it's the main reason why we link people to this file

.github/CONTRIBUTING.md
```diff
@@ -273,403 +172,67 @@ There are two approaches to creating a Pull Request: via terminal and via Visual
...
 
-3. Откройте [TerraFirmaGreg-Core-Modern] и скачайте последний релиз.
-4. В папке mods найдите jar-файл TerraFirmaGreg-Core-Modern, удалите его и замените на только что скачанный.
 
->[!TIP]
-> Более новые версии могут быть доступны в Github Actions. 
-> Кроме того, если вы разрабатываете [TerraFirmaGreg-Core-Modern], вы можете скопировать собранные jar-файлы для их тестирования.
+1. Create a new Prism Launcher instance with minecraft version 1.20.1 and forge.
+2. Get the latest version of [Modpack-Modern](https://github.com/TerraFirmaGreg-Team/Modpack-Modern)
```
Pyritie: instead, I would suggest they download the latest build from the actions tab, since that will also include mods and so on

.github/CONTRIBUTING.md
```diff
@@ -273,403 +172,67 @@ There are two approaches to creating a Pull Request: via terminal and via Visual
...
-  ```bash
-  # Форматировать все файлы
-  npm run format --prefix kubejs
-
-  # Проверка форматирования без изменений
-  npm run format:check --prefix kubejs
-  ```
+1. Set up a dev instance for Prism Launcher.
+2. Make forks of Modpack-Modern, Core-Modern and Tools-Modern and clone them.
+3. symlink your Modpack-Modern's configs and kubejs with your prism instance.
```
Pyritie: this works for windows but not other OSes. I suggest linking to the symlink wikipedia article or something instead. Also, don't symlink your configs folder, minecraft fails to write to it and will cause lots of problems

SakuraKitsurugi (changes requested):
Do not remove the Russian part, just ask for someone else to update it later.

.github/CONTRIBUTING.md
```diff
@@ -273,403 +172,67 @@ There are two approaches to creating a Pull Request: via terminal and via Visual
...
+    - `mklink /D C:\Users\user\AppData\Roaming\PrismLauncher\instances\INSTANCE\minecraft\kubejs C:\Users\user\tfg\Modpack-Modern\kubejs`
+4. Test your changes by building your core jar, adding it to your instance and launching it.
 
-  **Проверка типов TypeScript:**
+If you're using vscode, you may want to try [ProbeJs](https://marketplace.visualstudio.com/items/?itemName=Prunoideae.probejs) for kubejs changes (optional).
```
Pyritie: there's no guide on how to set up intellij to work with tfg-core

Pyritie (changes requested):
I don't know why you felt the need to completely remove all the formatting that the version used to have, Can you add it back? It went into more detail on some steps instead of just "clone the repo" which means nothing to people who are totally new, which this guide is supposed to be aimed at.

.github/CONTRIBUTING.md
ashleney: the one that takes 10 minutes to download? alright...

.github/CONTRIBUTING.md
ashleney: I don't use Intellij personally

PR #4114, opened 2026-05-24
Update for Fluid Logistics 1.0.5

Bumperdo09:
## What is the new behavior?
FluidLogistic 1.0.5 improves the UI of the stock ticker when displaying fluids considerably

It also adds a block, an infinite water source in 1 block, that this pack doesn't need because aqueous accumulators already provide infinite water. This PR removes it.


Pyritie:
could you put the new block (and the other removed ones) into a file [in here](https://github.com/TerraFirmaGreg-Team/Modpack-Modern/tree/dev/config/reliable_remover) instead? it'll also take care of removing their recipes and deleting the item from a player's inventory if they ever somehow get a hold of one.

Pyritie (approved):

Redeix (approved):

PR #4134, opened 2026-05-27
Treated Wood Recipes

### LINKED ISSUE #3957: [💎]: Creosote wood treatment recipes to treat items such as sticks, plank items, etc.

- [x] No, it was not.

Recipe for Treated Wood types from combining Creosote + Any untreated variant plank block/sticks.
Specifically listed are Treated Wood block/stick variants that have no Creosote related recipe:
Treated Wood Plank    (Item)
Treated Wood Long Stick
Treated Wood Stick
Treated Wood Frame
Treated Wood Slab
Treated Wood Fence
Treated Wood Pressure Plate
Treated Wood Trapdoor
Treated Wood Stairs
Treated Wood Button
Treated Wood Fence Gate
Treated Wood Door

### Why is this needed?

Block transition from standard variant planks to treated planks is iffy.  Most to all of the variant planks blocks cannot be sawed back down into planks, making this a difficult endeavor to transition from variant plank to treated plank, if you wish to decorate with said block.  Any previous variant plank blocks must be removed and replaced with the new treated planks through arduous crafting already visited previously.
In addition, any reason to use more Creosote early on is a welcome one.

Low Prio 100%

SakuraKitsurugi:
### What is the new behavior?
Adds more treatment wood recipes.

### Implementation Details
- Recipes are based on how much lumber is used.
- 1 Lumber is 1000 ticks and 25mB creosote.
- Buttons have an odd multiplier so I did not implement them.
- Frame, Window and Pane do not have regular wood variants so I did not implement them. The only items that are close are the compat ones, but they do not have tags.

### Additional Information
- Closes #3957
- Used this PR in the tutorial video.


Pyritie (dismissed):

Pyritie:
will need lang strings for all the recipe IDs too

Redeix (changes requested):
Probably should use proper item output IDs and use the linuxUnfucker() for the recipe id. That way it can support any future additions to the object not just gtceu items

SakuraKitsurugi:
the linuxUnfucker() is actually what i did at first lol, yeah will fix it

Pyritie (approved):

Redeix (approved):

PR #4182, opened 2026-06-04
Make a bunch of items replaceable by tracks

### LINKED ISSUE #4136: [💎]: Create tracks should be able to replace/remove sticks, twigs, small rocks and all similar ground elements when placing them

- [x] No, it was not.

When placing Create mod train tracks they should be able to replace all ground element:
i'll very likely miss some but this include 
- sticks
- twigs
- small rocks
- ore nuggets
- rotten flesh
- bones
- crystal buds and gregtech ore indicators
- all of these things but with snow
- very likely other things on other planets i forgot about

### Why is this needed?

It's so frustrating to place track and realizing a single stick on the ground made a hole in your railroad

ashleney:
fixes #4136

can_be_snow_piled seems to include everything listed in the issue except for afc and tfg twigs

this voids the replaced blocks

https://github.com/TerraFirmaGreg-Team/Core-Modern/pull/480

### DIFF
--- a/kubejs/server_scripts/create/tags.js
+++ b/kubejs/server_scripts/create/tags.js
@@ -191,7 +191,6 @@ const registerCreateBlockTags = (event) => {
 	event.add('tfc:no_icicle_generation', 'create:chain_conveyor')
 
     event.add('tfg:track_replaceable', '#tfc:can_be_snow_piled')
-    event.add('tfg:track_replaceable', '#tfc:twigs')
 }
 
 
--- a/kubejs/server_scripts/create/tags.js
+++ b/kubejs/server_scripts/create/tags.js
@@ -189,6 +189,9 @@ const registerCreateBlockTags = (event) => {
 	event.add('create:non_movable', 'tfg:mafic_hornfels')
 
 	event.add('tfc:no_icicle_generation', 'create:chain_conveyor')
+
+    event.add('tfg:track_replaceable', '#tfc:can_be_snow_piled')
+    event.add('tfg:track_replaceable', '#tfc:twigs')
 }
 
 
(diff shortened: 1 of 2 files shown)


Pyritie:
hm those twigs should also have that tag...

Pyritie:
out of curiosity, how does create normally handle this for things like grasses? I'm guessing it checks #minecraft:replaceable or something?

Pyritie (dismissed):

Pyritie:
it'd be good to add these twigs to the can be snow piled tag anyway. Out of curiosity how does this interact with `minecraft:snow` larger than size 1? what about `tfc:snow_pile`? might be good to add those to your new tag too

ashleney:
Base create checks BlockState::canBeReplaced and #minecraft:flowers. I frankly do not know what minecraft's BlockState::canBeReplaced checks but I don't think it's #minecraft:replaceable considering loose rocks were replaceable and yet tracks did not replace them.

It handles snow and snow piles fine, though there's the issue of *inidividual* tracks being placed on top of snow instead of replacing the snow but that is an issue for everything, not just tracks.

kubejs/server_scripts/create/tags.js
```diff
@@ -189,6 +189,9 @@ const registerCreateBlockTags = (event) => {
 	event.add('create:non_movable', 'tfg:mafic_hornfels')
 
 	event.add('tfc:no_icicle_generation', 'create:chain_conveyor')
+
+    event.add('tfg:track_replaceable', '#tfc:can_be_snow_piled')
+    event.add('tfg:track_replaceable', '#tfc:twigs')
```
Pyritie: is the twigs still needed here? I think we might be missing the ginkgo twig as well? or is that already in the tag

ashleney: ginkgo doesn't have a twig, dunno if that's intended

Pyritie: oh fr? I'll have to add that

Pyritie (approved):

TomPlop (approved):

PR #4219, opened 2026-06-09
Add alternate rubbers to hazmat armor pieces

lickorice:
## What is the new behavior?
Allows alternate rubbers (silicone, styrene butadiene) to be used in making of hazmat pieces. Stumbled upon a minor QoL problem where I don't have the raw, plant-based (latex) rubber when I needed to make a set. This should solve that, and we already have precedents with conveyor belts and pumps being allowed to be crafted using alternative rubber materials.

## Implementation Details
Simply adds an `event.recipes.gtceu.assembler` hook for-block in `gregtech/recipes.js`.

## Outcome
Adds new `gtceu:assembler/hazmat_[chestpiece | leggings | boots]_[silicone_rubber | styrene_butadiene_rubber]` recipe IDs using `event.recipes.gtceu.assembler`

## Additional Information
Ensured it matches the existing recipe using rubber (10 * 20 ticks, at 0.94A LV)
- Chestpiece
image

- Leggings
image

- Boots
image

Checked if a configuration circuit is needed -- doesn't seem to conflict with existing assembler recipes for silicone / styrene butadiene rubber so I'm deferring its usage for now.

Discord:
lickorice
319285994253975553

### DIFF
--- a/kubejs/server_scripts/gregtech/recipes.js
+++ b/kubejs/server_scripts/gregtech/recipes.js
@@ -455,5 +455,28 @@ const registerGTCEURecipes = (event) => {
 
 	event.replaceInput({ id: 'gtceu:shaped/cracking_unit' }, '#gtceu:circuits/hv', '#gtceu:circuits/ev')
 
+  // Allow alternate rubbers for hazmat pieces
+
+  const ALTERNATE_RUBBERS = [
+		'silicone_rubber',
+		'styrene_butadiene_rubber',
+	]
+  ALTERNATE_RUBBERS.forEach(material => {	
+    event.recipes.gtceu.assembler(`gtceu:hazmat_chestpiece_${material}`)
+      .itemInputs(`2x #forge:plates/${material}`, '7x gtceu:polyvinyl_chloride_plate', '3x #forge:plates/lead')
+      .itemOutputs('gtceu:hazmat_chestpiece')
+      .duration(10*20)
+      .EUt(GTValues.VA[GTValues.LV])
+    event.recipes.gtceu.assembler(`gtceu:hazmat_leggings_${material}`)
+      .itemInputs(`2x #forge:plates/${material}`, '5x gtceu:polyvinyl_chloride_plate', '2x #forge:rods/iron', '4x #forge:rings/steel')
+      .itemOutputs('gtceu:hazmat_leggings')
+      .duration(10*20)
+      .EUt(GTValues.VA[GTValues.LV])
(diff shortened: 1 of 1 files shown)


lickorice:
Would have done #tfg:rubber_plates from the get-go, but that involves removing the default (? i think?) recipe for the pieces, and it already meshes well with the other existing gtceu stuff (conveyors, pumps) where the same rubber type is always used

Pyritie:
would doing `event.replaceInput()` on the assembler recipes work? Replacing `#forge:plates/rubber` with `#tfg:rubber_plates`?

lickorice:
Lemme try now

lickorice:
works perfectly, updated the PR.
image

on an unrelated note -- can we do the same thing for conveyors and pumps? just worried that it might break AE2 recipes somehow or there's some side effect i'm not aware of (or a design decision) if i do that.

(if yes, ill do that on a diff PR)

thanks, pyritie!

Pyritie (approved):

Pyritie:
> on an unrelated note -- can we do the same thing for conveyors and pumps? just worried that it might break AE2 recipes somehow or there's some side effect i'm not aware of (or a design decision) if i do that.

conveyors and pumps should already have recipes for the different rubbers

TomPlop:
I wonder if these shouldn't be part of GT rather

lickorice:
> > on an unrelated note -- can we do the same thing for conveyors and pumps? just worried that it might break AE2 recipes somehow or there's some side effect i'm not aware of (or a design decision) if i do that.
> 
> conveyors and pumps should already have recipes for the different rubbers

yup but they don't allow mixing right

image

TomPlop:
> > > on an unrelated note -- can we do the same thing for conveyors and pumps? just worried that it might break AE2 recipes somehow or there's some side effect i'm not aware of (or a design decision) if i do that.
> > 
> > 
> > conveyors and pumps should already have recipes for the different rubbers
> 
> yup but they don't allow mixing right
> 
> image

and shouldn't

lickorice:
do i revert to the non-mixed (non-`#tfg:rubber_plates`) setup to match design decisions or do i keep current implem? thanks!

Pyritie:
leave your PR as it is

Redeix (approved):

TomPlop:
I wonder if I didn't prefer the first iteration that follows the other components logic?

Pyritie:
doesn't really matter, we have other things using this tag already

PR #4253, opened 2026-06-13
fix server launch flags

Th1nk-Wave:
## What is the new behavior?
start_server.sh and start_server.bat use the existing JVM memory flags as expected

### DIFF
--- a/.pakku/server-overrides/start_server.bat
+++ b/.pakku/server-overrides/start_server.bat
@@ -1,3 +1,3 @@
 @echo off
-java -jar -Xmx6024M -Xms1024M minecraft_server.jar nogui
+java -Xmx6024M -Xms1024M -jar minecraft_server.jar nogui
 pause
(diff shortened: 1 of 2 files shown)


.pakku/server-overrides/start_server.bat
```diff
@@ -1,3 +1,3 @@
 @echo off
-java -jar minecraft_server.jar -Xmx6024M -Xms1024M nogui
```
Xikaro, line 2: The -jar flag should accept the jar file. After the transfer, you left it in place, fix it.

.pakku/server-overrides/start_server.sh
```diff
@@ -1,2 +1,2 @@
 #!/bin/bash
-java -jar minecraft_server.jar -Xmx6024M -Xms1024M nogui
```
Xikaro, line 2: The -jar flag should accept the jar file. After the transfer, you left it in place, fix it.

Xikaro (changes requested):

PR #4265, opened 2026-06-15
Restored recipes for copying configured filters

Robak132:
## What is the new behavior?
Restored create recipes to copy configuration of filters in crafting grid.

Filled filter + Empty filter = 2x Filled filter

None

Discord nickname: Robak132

### DIFF
--- a/kubejs/server_scripts/create/recipes.js
+++ b/kubejs/server_scripts/create/recipes.js
@@ -5,7 +5,7 @@ const registerCreateRecipes = (event) => {
 
 	// Удаление рецептов мода create 
 	event.remove({
-		not: [			
+		not: [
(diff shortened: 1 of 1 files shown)


Pyritie (commented):
is this something that instead could be added to the block at the top of the file? if you add create's original recipe IDs to it, it'll preserve them

Robak132:
I tested and it didn't work when in this large block on top.

Pyritie:
I mean the original recipe IDs from create, not your custom one

Robak132:
Sth like that?

Pyritie (commented):
does it work?

Robak132:
Yep, before i tried with id selector instead of type so that's why it didn't

Pyritie (approved):

Redeix (approved):

PR #4312, opened 2026-06-20
Chromium changes + allow nether quartz for SiC ingots

Pyritie:
- allow nether quartz for SiC ingots, with exactly the same recipes as regular quartz. NQ is basically useless at the moment so this seems fitting (what do you think about allowing opal as well, since it's also just SiO2?)

- change some of the chromium EBF recipes to cupronickel coils instead of kanthal, because you need chromium to make kanthal in the first place, and the *only* pre-kanthal recipe currently is ruby slurry -> chromium 3 oxide -> EBF with aluminium, which feels incredibly bad when you need a whole bunch of chromium for your coils.

(also, since chromium for the kanthal coils is now much more of a pain, what do you think about making another part of it a little easier, to compensate? like making some of the other SiC or rené EBF recipes faster, since a lot of making these coils is EBF stuff now)


Pyritie:
image

also, is it intended that there's no yellow garnet ore? it's only on the moon, and at that point you have chromite

kubejs/server_scripts/tfg/ores_and_materials/recipes.chromium.js
```diff
@@ -51,7 +51,7 @@ function registerTFGChromiumProcessing(event) {
                 .outputFluids(Fluid.of('gtceu:carbon_monoxide', 3000))
                 .duration(36 * 20)
                 .EUt(GTValues.VA[GTValues.MV])
-                .blastFurnaceTemp(2050)
+                .blastFurnaceTemp(1250)
```
TomPlop: That one is a big no because it's exactly intended for the player to use Aluminium dust at first with this path. You unlock the loop after when you made your first Kanthal. The alternative is to use Chromite dust that you can get through ore processing and Calcium Carbonate.

kubejs/server_scripts/tfg/ores_and_materials/recipes.chromium.js
```diff
@@ -92,7 +92,7 @@ function registerTFGChromiumProcessing(event) {
                 .itemOutputs(Item.of('tfg:sodium_chromate_dust', 14), Item.of('tfg:wollastonite_dust', 15))
                 .duration(32 * 20)
                 .EUt(GTValues.VA[GTValues.MV])
-                .blastFurnaceTemp(1850)
+                .blastFurnaceTemp(1250)
```
TomPlop, line 95: yeah that one should be cupronickel

TomPlop (approved):

TomPlop:
Chromium_full

TomPlop:
> image
> also, is it intended that there's no yellow garnet ore? it's only on the moon, and at that point you have chromite

Yeah you just get some from garnet sand and it had better yield from the moon before the changes. We could add more garnet sand veins on Earth as it's pretty low right now. You wanted more sandy veins?

TomPlop:
Sounds good for Nether Quartz I guess it's easier for Beneath players.

TomPlop (changes requested):
That one is a big no because it's exactly intended for the player to use Aluminium dust at first with this path. You unlock the loop after when you made your first Kanthal. The alternative is to use Chromite dust that you can get through ore processing and Calcium Carbonate.

Pyritie:
> Yeah you just get some from garnet sand and it had better yield from the moon before the changes. 

oh true, you can get yellow garnet from sphalerite and grossular oreproc too.

> We could add more garnet sand veins on Earth as it's pretty low right now. You wanted more sandy veins?

both of the mineral sands veins only have 7% garnet sands, we could bump that number up

TomPlop (approved):

PR #4335, opened 2026-06-22
Add shaped crafting recipes for train tracks

### LINKED ISSUE #4292: [💎]: Add crafting table recipes for rails

- [x] No, it was not.

Add crafting table recipes for crafting train tracks.

### Why is this needed?

Everything else for crafting train stuff is steel age, but the tracks themselves require deployers, which are halfway through the steam age.

We don't really have anything else that would be suitable for crafting them earlier, so a (bad) crafting table recipe would be an ok substitute.

strangelyng:
## What is the new behavior?
Adds shaped crafting recipes for rails.

## Implementation Details
Normal tracks
normal_track
Wide tracks
wide_track
Narrow tracks
narrow_track

Monorail
monorail

Stone variants use the `#tfg:rock_slabs` and `#forge:stone` tags.

## Outcome
Closes: #4292

**Discord:** strangelyng


SakuraKitsurugi (changes requested):
Wide tracks shaped recipes are cheaper wood-wise than other recipes. Recommend switching S and B in shaped recipes.

Also, might just be a nitpick and is not really an issue, but usually (or I guess at least me) the shaped letters used go in alphabetical order (A, B, C, D...)

strangelyng:
I made the requested changes, thanks for catching that! I also adjusted the narrow tracks to only use a single slab, to match the inputs for the sequenced assembly recipes as well. Seems like a weird inconsistency otherwise, mostly the nerf is just in using tool durability and halved yields.

Pyritie (approved):

SakuraKitsurugi (approved):

PR #4360, opened 2026-06-24
Fix purified tin and silver producing blacklisted TFC ingots

### LINKED ISSUE #4359: [🐛]: Purified tin and silver ores produce blacklisted TFC ingots when smelted in furnace

- [x] No, it was not.

### Computer OS

Windows 10

### Modpack Version Used

0.13.0 and dev branch

### Minecraft Environment

Singleplayer

### New Worlds

Yes

Unnecessary

### Expected Behavior

Expected to properly produce silver and tin ingots from purified silver and tin ores respectively.

### Actual Behavior

Smelting purified silver/tin produces TerraFirmaCraft ingots, which are blacklisted by Reliable Remover and thus deleted from the inventory, instead of producing the proper GregTech ingots. In addition, the smelting recipe for smelting purified silver/tin is hidden in EMI, Reliable Remover probably hides it as it produces a blacklisted item.

https://github.com/user-attachments/assets/89990f40-bd5a-4bc8-993b-5e68c6442599

### Steps to Reproduce

Produce "Purified Tin Ore" (create:crushed_raw_tin) via spawning, or via spawning any form of raw tin and crushing, then washing it. "Purified Tin Ore" is obtainable in survival. Then, place the "Purified Tin Ore" in a furnace. This produces TerraFirmaCraft Tin Ingots, which are blacklisted and thus are deleted from the machine inventory if exited or from the player's inventory if retrieved from the machine. The same is true for "Purified Silver Ore" (create:crushed_raw_silver), sourced from crushing and washing raw native silver.

### Additional Information

This seems to be a problem with the create "purified" ores. This does not affect purified gold, copper, or lead ores (all obtainable in survival) (even though they are also sourced from Create) as they do not have a TFC ingot equivalent to create. This does affect "Purified Zinc Ore", but this is not obtainable in survival so not a priority. As a sidenote, "Purified Zinc Ore" appears in EMI, even though it is unobtainable, which should probably be fixed.

I tested on both a Prism install of 0.13.0 and a dev instance (following the contribution instructions) and the error occurs on both.

CaptainGold1:
## What is the new behavior?
Relevant issue: #4359 
Purified silver and tin now produce correct GT ingots instead of blacklisted TFC ones.

## Implementation Details
Changes the furnace recipes for purified silver and tin to explicitly produce gtceu:silver_ingot/tin_ingot instead of referencing the forge ingot tag.

## Outcome
Fixes: #4359 

## Additional Information
I cannot be 100% certain this bug is not due to some weird issue with my installation, though it shouldn't be as I have tried on two separate fresh installs. **This should be replicated first before merging**.

I am unsure of the ramifications of switching the furnace recipe over to explicitly provide GT ingots, e.g. if the ingot is changed in the future to another mod's ingot, it would not properly switch over. I am personally unsure if there is another way to fix this issue, ideally the recipe would still reference the tag but the tag would properly exclude the TFC ingots.

Discord username: CaptainGold1

### DIFF
--- a/kubejs/server_scripts/create/recipes.js
+++ b/kubejs/server_scripts/create/recipes.js
@@ -2421,8 +2421,8 @@ const registerCreateRecipes = (event) => {
 	event.smelting('#forge:ingots/copper', 'create:crushed_raw_copper')
 	event.smelting('#forge:ingots/gold', 'create:crushed_raw_gold')
 	event.smelting('#forge:ingots/zinc', 'create:crushed_raw_zinc')
-	event.smelting('#forge:ingots/silver', 'create:crushed_raw_silver')
-	event.smelting('#forge:ingots/tin', 'create:crushed_raw_tin')
+	event.smelting('gtceu:silver_ingot', 'create:crushed_raw_silver')
+	event.smelting('gtceu:tin_ingot', 'create:crushed_raw_tin')
 	event.smelting('#forge:ingots/lead', 'create:crushed_raw_lead')
 
 }


Pyritie:
confirmed

Pyritie (approved):
something we should keep an eye on for sure

Redeix (approved):

PR #4368, opened 2026-06-25
Fix Channel Casting Clay Recycling

TreyR9:
Quick fix because the recycling recipe for unfired channel casting items was broken. Removed the TFCChannelCasting tags and replaced them with TFC tags.


Redeix:
Are these tags not from us? TFCChannelCasting has recycling?

TreyR9:
Couldn't find it added in the repo, and it doesn't work in-game. The TFC recycling tag works while the TFCChannelCasting does not.

Pyritie:
maybe the original mod forgot to implement the recycling recipes. I think firmalife did too

Pyritie (approved):

Redeix (approved):

PR #4384, opened 2026-06-26
Ice slush cooling

theMegaTech:
Added ice slush cooling as like, duh, cool water should cool better than not cool water

Made it slower than olive oil as olive oil is way more effort than bothering with some ice, but still plenty fast

### DIFF
--- a/kubejs/server_scripts/tfc/recipes.js
+++ b/kubejs/server_scripts/tfc/recipes.js
@@ -334,4 +334,29 @@ const registerTFCRecipes = (event) => {
 		.id("tfg:quern/candles")
 	event.recipes.tfc.quern("gtceu:tiny_wax_dust", "gtceu:wax_nugget")
 		.id("tfg:quern/wax_nugget")
+
+	// Ice Slush cooling
+	event.custom({
+		"type": "tfc:barrel_instant",
+		"input_item": {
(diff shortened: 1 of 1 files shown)


Pyritie (commented):
would it be possible to write this in the kubejs-tfc kind of way instead, for consistency?
https://notenoughmail.github.io/kubejs_tfc/1.20.1/recipes/#barrel-instant
https://notenoughmail.github.io/kubejs_tfc/1.18.2/ingredients/#heatable
https://notenoughmail.github.io/kubejs_tfc/1.20.1/bindings/isp/

theMegaTech:
Apparently its possible, didn't try initially because had problems with barrel instant before (cheese rot stuff)

Tested this one, does work the same

Pyritie (dismissed):

Redeix (dismissed):
Before I merge do you think there's any other fluids that should be added? Or a fluid tag?

Pyritie:
hmm... I know olive oil works but I'm not sure if it also allows any oil.
There's also generic tags for clean water and any water, if we want

Redeix (dismissed):

Redeix:
The IDs all have slush in them now. But I don't think thats important enough to fix

Pyritie (approved):

PR #4408, opened 2026-06-30
wool block loom & assembler parity

Froffy025:
## What is the new behavior?
Adds loom recipes for linen cloth and silk cloth to wool for parity with the assembler

## Implementation Details
You might be able to macerate linen wool into some wool string and get wool cloth from linen if you are insane

I'm assuming the loom doesn't allow for item tags because nothing else uses them but i don't know

## Additional Information
image

froffy


SakuraKitsurugi:
Honestly, in my opinion, wool blocks should only be made from wool cloth.

TomPlop:
Agree get some animals, the assembler recipes exist more as a progression step and ease of life.

Froffy025:
> Agree get some animals, the assembler recipes exist more as a progression step and ease of life.

at the current conversion ratio, you can get 1 wool block per animal per shearing, and can only do it in batches of 8. using musk ox, who grow wool the fastest, that's 1 wool block every 4 days per animal. it gets worse with alpaca (6 days/wool) and sheep (9 days/wool). that's abysmal, even by pre-industrial standards.

especially since we have linen, which i don't think has many uses outside of being #forge:light_sewing_cloth. why not give players the option to build with it?

SakuraKitsurugi:
Hmm, maybe silk cloth then, but definitely not linen cloth.

Pyritie:
why don't we just rename them to cloth blocks? they're only used for decoration

Froffy025:
> why don't we just rename them to cloth blocks? they're only used for decoration

i'd like this solution.. i've got like 50 alpacas begging for mercy at the moment >.<

SakuraKitsurugi:
@Froffy025 if Pyritie is fine with it then you can make a PR to the Tools repo to rename it :+1:

SakuraKitsurugi (approved):

Redeix (approved):

PR #4415, opened 2026-07-01
tinderbox

Froffy025:
## What is the new behavior?
Adds an iron-age version of the flint and steel, the tinderbox.

## Implementation Details
Otzi had something like this. His fire striker was a piece of pyrite. We will be using the high carbon content of a cast iron plate as our fire striker, and a bundle of charcloth as our tinder.

## Outcome
Iron age QOL since that's when you use the forge the most. Lasts much longer than a matchbox, but cannot be refilled. It's more expensive than flint and steel, while being less durable - so if you're still lighting a lot of fires after getting a blast furnace, it makes sense to switch.

## Additional Information
image


Froffy025:
if the overlap with matches is too much, i'd also be down to change it to work like upgraded firesticks, if i can figure out how!

SakuraKitsurugi:
Would be more interesting if it worked like the fire starter and require being held down for some time.

TomPlop:
Cast iron sounds really easy but the fire starter is also really boring anyway so.

I would be more open minded if it had durability and a random chance to fully break.

SakuraKitsurugi:
It does have 64 durability look at the code, but the random chance of fully breaking might be too annoying.

Pyritie:
why close?

Froffy025:
took one look at tfc firestarter code and couldn't comprehend the horrors... (dogshit coder) (mixins scare me) (got frustrated)

feel free to use the sprite if you like the idea 😅 i can't stomach looking at the IDE for a little while

PR #4488, opened 2026-07-09
Tower Revamp

Filiipa:
## What is the new behavior?
_Adds new variations to the tower, based on the new gilsonite bricks, basalt, and moss variants._

## Implementation Details
_I overcomplicated a lot, this is so it's modular and can be changed at a blink of an eye, if i do it of course, most people will be lost in the first 30 minutes._

## Outcome
_Removes the top and bottom tower structures, making the tower a whole structure._


Redeix (approved):
Hmmm yeah I don't understand any of this. LGTM 👍

Pyritie (approved):
damn this shit crazy

I mean if you can make a new world, tp to beneath, find towers and they all have 5 segments, and their loot tables etc all look correct and the segments all connect, then it's good to me too

PR #4531, opened 2026-07-15
fix: missing tags of coconut milk

CresLeaf:
## What is the new behavior?
Re-add tags
- `tfc:milks`
- `tfc:drinkables`
- `tfc:any_drinkables`

to `firmalife:coconut_milk `

Add `tfg:nonplaceable` to `firmalife:bucket/coconut_milk`

None, except balancing around the new crops

crescent1a

### DIFF
--- a/kubejs/server_scripts/firmalife/tags.js
+++ b/kubejs/server_scripts/firmalife/tags.js
@@ -2,8 +2,8 @@
 "use strict";
 
 const registerFirmaLifeItemTags = (event) => {
-    
-   
+
+
     // Make our own "dried fruit" tag so we can display something in EMI - used for yeast starter
     global.FOOD_FRUIT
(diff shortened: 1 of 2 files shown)


Redeix:
these were all removed on purpose. Fluids cant expire. Coconut milk is intentionally a worse source of dairy compared to farm animals. The only reason I didnt out-right remove the fluid was in case we wanted to use it for recipes later. Also you need to control your linter from making so many redundant line changes

Redeix:
Also, re-adding the tags would have solved nothing since coconut milk doesnt even have food data

CresLeaf:
> Also, re-adding the tags would have solved nothing since coconut milk doesnt even have food data

it does add player's dairy nutrient when drunk by player, though i don't figure out why

PR #4562, opened 2026-07-20
peat bricks

applenper:
an attempt to make peat a little more useful
##
peat can now be dried (or coked) into bricks

image
image

##

peat bricks do not burn hotter than regular peat but can be used in a charcoal forge or burn for 45 seconds in a boiler

image

image

##

when used with a bellows, peat bricks cannot melt bronze or copper (maxing out at orange) but can still bring it to a workable/weldable temperature

this makes them useful for working bronze without expending charcoal

image


kubejs/server_scripts/tfg/primitive/recipes.peat.js
Pyritie, line 1: Change this to the KubeJS-TFC format instead please -- and I don't think this needs its own file either. Is there a file for dirt/mud related recipes? or maybe you could put this next to the peat coke oven recipe?

Pyritie (changes requested):
don't forget the tools-modern PR for the langs too

kubejs/server_scripts/tfg/primitive/recipes.peat.js
Pyritie: sorry I thought I pasted the link earlier https://notenoughmail.github.io/kubejs_tfc/1.20.1/recipes/#firmalife-drying

Pyritie (approved):

Redeix (approved):
I feal like "briquette" would be a better translation since it will prevent people from thinking its a building material. And briquette is classically used for dried/compressed solid fuels

PR #4582, opened 2026-07-22
New steam engines

Pyritie:
So we're using a resource pack (the default models don't look great) but I can't just overwrite textures from the resource pack to add the greate tier colors. So I have to copy the models out of it too.
The resource pack is getting cursebux either way, so this should be ok.

https://github.com/TerraFirmaGreg-Team/Tools-Modern/pull/505
https://github.com/TerraFirmaGreg-Team/Core-Modern/pull/535


kubejs/server_scripts/create_steam_powered/recipes.js
```diff
@@ -0,0 +1,123 @@
+// priority: 0
+"use strict";
+
+function registerCreateSteamPoweredRecipes(event) {
```
Redeix, line 4: Shouldnt these all have assembler recipes too since they use tools?

Pyritie:
also I forgot the pakku changes

Redeix (approved):

PR #4589, opened 2026-07-24
Feature: added electrolyzer recipes for quartz and smoky quartz dusts

### LINKED ISSUE #4539: [💎]: add electrolysis recipes for quartz and smoky quartz dusts

- [x] No, it was not.

allow Quartz Dust and Smoky Quartz Dust to be electrolyzed for silicon and oxygen

### Why is this needed?

because silicon and oxygen are already very common resources, smoky quartz currently has very little use, and regular silicon dioxide dust can be electrolyzed

hollow-earth:
## What is the new behavior?
Added a recipe to electrolyze 3x quartz dust/smoky quartz dust into 1x silicon dust and 2B oxygen gas

## Implementation Details
Following the example set by rose quartz, I made both recipes 60 EU/t, 7 seconds. Could be altered if needed

## Outcome
Fixes #4539

### DIFF
--- a/kubejs/server_scripts/tfg/ores_and_materials/recipes.quartzes.js
+++ b/kubejs/server_scripts/tfg/ores_and_materials/recipes.quartzes.js
@@ -22,7 +22,21 @@ function registerTFGQuartzRecipes(event) {
 		.itemOutputs('1x #forge:dusts/nether_quartz', '8x minecraft:redstone')
 		.duration(7 * 20)
 		.EUt(60)
-	//#endregion
+
+	event.recipes.gtceu.electrolyzer('electrolyze_smoky_quartz_dust')
+		.itemInputs('3x #forge:dusts/nether_quartz')
+		.itemOutputs('gtceu:silicon_dust')
+		.outputFluids(Fluid.of('gtceu:oxygen', 2000))
+		.duration(7 * 20)
+		.EUt(60)
+
+	event.recipes.gtceu.electrolyzer('electrolyze_quartz_dust')
+		.itemInputs('3x #forge:dusts/quartzite')
+		.itemOutputs('gtceu:silicon_dust')
+		.outputFluids(Fluid.of('gtceu:oxygen', 2000))
+		.duration(7 * 20)
+		.EUt(60)
+  //#endregion
 
 	//#region Glass
 
--- a/kubejs/server_scripts/tfg/ores_and_materials/recipes.quartzes.js
+++ b/kubejs/server_scripts/tfg/ores_and_materials/recipes.quartzes.js
@@ -23,14 +23,14 @@ function registerTFGQuartzRecipes(event) {
 		.duration(7 * 20)
 		.EUt(60)
 
-	event.recipes.gtceu.electrolyzer('electrolyze_smoky_quartz_dust')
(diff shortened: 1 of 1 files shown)


kubejs/server_scripts/tfg/ores_and_materials/recipes.quartzes.js
```diff
@@ -22,7 +22,21 @@ function registerTFGQuartzRecipes(event) {
 		.itemOutputs('1x #forge:dusts/nether_quartz', '8x minecraft:redstone')
 		.duration(7 * 20)
 		.EUt(60)
-	//#endregion
+
```
Redeix, line 25: Recipe IDs need to be in the `tfg:` namespace. Also I thought there was a material flag to enable decomposition

TomPlop:
This isn't the way to do it and why? That sounds more like a trap that an actual good recipe to add lol

TomPlop:
This isn't the way to do it and why? That sounds more like a trap that an actual good recipe to add lol

kubejs/server_scripts/tfg/ores_and_materials/recipes.quartzes.js
hollow-earth: You mean something like `tfg:electrolyze_smoky_quartz_dust`?

hollow-earth:
> This isn't the way to do it and why? That sounds more like a trap that an actual good recipe to add lol

Followed the recipe from silicon dioxide dust, which is obtained from 8x sand. Decomposes into 1x silicon and 2B oxygen as well. If you think it's a bad idea we can just close the PR

Pyritie:
time to learn about gregtech's material system!

https://gregtechceu.github.io/GregTech-Modern/1.20.1/Modpacks/Materials-and-Elements/Material-Flags/

you can use the `material_modification.js` files to add flags to materials. GT uses these flags (and properties) to then generate recipes and other things

hollow-earth:
> time to learn about gregtech's material system!
> 
> https://gregtechceu.github.io/GregTech-Modern/1.20.1/Modpacks/Materials-and-Elements/Material-Flags/
> 
> you can use the `material_modification.js` files to add flags to materials. GT uses these flags (and properties) to then generate recipes and other things

Doesn't look like there's a way to remove built in tags... at least not easily 
https://github.com/GregTechCEu/GregTech-Modern/blob/ee635d636e87e06ae827de9b4eb98e840a394685/src/main/java/com/gregtechceu/gtceu/api/data/chemical/material/info/MaterialFlags.java

kubejs/server_scripts/tfg/ores_and_materials/recipes.quartzes.js
hollow-earth: Yeah, after investigating it seems that there's a `GTMaterialFlags.DECOMPOSITION_BY_ELECTROLYZING` material tag. The issue is that `GTMaterialFlags.DISABLE_DECOMPOSITION` is present by default and seems to override the former.

Pyritie: ok, just put these in the `tfg:` namespace then

Pyritie (approved):

PR #4625, opened 2026-07-29
Compost Tumbler

Redeix:
- Re-added compost tumbler (with textures from vexxed visuals)
- Added assembler recipes for sweepers and pickers.
- Changed stack size of compost bundles from 16 to 32

> The field guide entry will be good enough for now. Ill re-write it and the quern and bellows entries in the patchouli rework branch. Ill probably add/update quests too.

Needed for https://github.com/TerraFirmaGreg-Team/Core-Modern/pull/539


SakuraKitsurugi (approved):

kubejs/server_scripts/firmalife/recipes.js
```diff
@@ -372,7 +385,38 @@ const registerFirmaLifeRecipes = (event) => {
...
+	// Compost Tumbler
+	event.remove('firmalife:crafting/compost_tumbler');
+
+	event.recipes.gtceu.shaped('firmalife:compost_tumbler', [
+		'BBE',
+		'CAD',
+		'BB '
+	], {
+		A: 'tfc:composter',
+		B: 'createdeco:andesite_bars',
```
Pyritie, line 406: intentionally specifically these? or would any bars be ok? since tin is harder to get now

Pyritie (approved):

kubejs/server_scripts/firmalife/recipes.js
Redeix: I went with tin alloy specifically just to match the color of the model and because it's still pre wrought iron. But I can make it use any bars if you prefer

Pyritie: I'm fine either way

PR #4633, opened 2026-07-30
Adding a quest for barrel racks

abculatter2:
Py suggested I do this, I think I put the thing in the right place?

## What is the new behavior?
Adding a new quest for barrel racks to the quest book, since it seems like a feature that's easy to miss if you don't read the guidebook fastidiously.

## Implementation Details
I think I did things right?? The file looks pretty straightforward.

## Additional Information

abculatter_2


Pyritie:
oh I meant just an optional task and not a separate quest, though this works too. Make sure they're both optional please and put the language strings over here, otherwise they'll get overwritten: https://github.com/TerraFirmaGreg-Team/Tools-Modern/tree/dev/LanguageMerger/LanguageFiles/tfg/en_us/Quests

Look at existing quests for how to set them up. You can also have the quest link directly to the field guide too

abculatter2:
For adding it to the other repo, do I just add a separate pull request over there as well? Also, I wasn't sure how to do any of the fancy stuff like adding a link (Or, apparently, how to add a line break instead of a page break...)

Pyritie:
yes, and you can look at other quests to see how to do those

abculatter2:
Hokay, that's all that I have the brainpower to muster tonight, so any additional alternations or issue fixing will have to be done by someone else or will have to wait until I'm not tired from work.

Pyritie (changes requested):
can you make these optional

Pyritie (approved):

PR #4648, opened 2026-07-31
Stone Dust Sinks

Redeix:
Adds a couple more sinks for stone dusts since I have way too much. Intentionally not that efficient.


Pyritie:
yeah go on then

PR #4716, opened 2026-08-08
Changed silk clothing repair recipe to use tags

olavfla:
Allows for repair of silk clothing with items with the tfg:lightweight_cloth tag which matches the material used to make it.

## What is the new behavior?
Changed the material repair recipe for silk clothing to match the materials used in creation. Allows repair with phantom silk and polycaprolactam fabric in addition to regular silk. 

## Implementation Details
Only a simple json change in kubejs/server_scripts/tfg/equipment/recipes.repair.js. Moved silk repair from SPECIAL_REPAIRS to SPACIAL_TAG_REPAIRS

**Discord name: Kvarok**


olavfla:
I didn't make an issue for this. Feel free to dismiss if you think it's a bad change.

Pyritie (approved):
sounds reasonable to me

PR #4760, opened 2026-08-14
Fixed some fence tags for Pastoral Rancher

Vazde:
## What is the new behavior?
Some fence items are missing block tags, and can't be used in multiblocks. This adds them.

## Outcome
Macaw's TFC fences can now be used with the Pastoral Rancher.


Pyritie (commented):
were there any other fences from other mods that were missing these tags?

Vazde:
> were there any other fences from other mods that were missing these tags?

Sorry, didn't check. But at least afterwards Ad Astra and AFC fences have the correct tag. Their (and MCW's) gates are missing the forge tags thought. So assumedly this fixes them too. But this change didn't account for gates, so I'll add that in a moment.

Vazde:
It seems that this change is incomplete. At the time when block tags are being assigned item tags are not available. They are only available during a reload (:

But it does seem that Macaw's fences and gates at last have the minecraft block tag, so could use that instead of item tags.

Pyritie:
a loop through wood types would solve this

Vazde:
I give up for now. It seems the problem is too complex for what I was ready for. Also: at least framed fence and bamboo fence are not "wooden".

PR #4730, opened 2026-08-09
Illage Paths Fix

Filiipa:
## What is the new behavior?
_Solves bug #4559 by making road blocks not generate on top of grass, "stomping" it instead._
_Linking an issue can be used alternatively to writing a description._

## Implementation Details
_Just adds a new processor to each path processor._

PR #4761, opened 2026-08-14
Fix to Path Fix

Filiipa:
## What is the new behavior?
_Fixes the previous fix (#4730)._

## Implementation Details
_Makes things not crash by not putting processors inside other processors._


Filiipa (commented):
should work now....

ashleney (approved):

Pyritie (approved):

PR #4766, opened 2026-08-15
Combustion engines

Pyritie:
recipes and data for the new combustion engines

also removed the flowing fluid compat because I fixed that bug

https://github.com/TerraFirmaGreg-Team/Core-Modern/pull/557


kubejs/server_scripts/tfg/machines/recipes.kinetic.js
```diff
@@ -0,0 +1,76 @@
+"use strict";
+
+function registerTFGKineticRecipes(event) {
+
+	event.shaped('tfg:generators/steel_combustion_engine', [
```
Redeix, line 5: No recycling?

Pyritie: good point

Redeix (approved):

PR #4804, opened 2026-08-20
Factory Jarred Pickles

Froffy025:
## What is the new behavior?
You can now create pickle jars in the food processor. This requires a lot more ingredients than other methods of vegetable processing. You'll need two cucumbers, a clove of garlic, one vegetable of your choice, one spice, four salt, and 250 mB of vinegar. The end result is a single, slightly more nutritious vegetable that will last for years.

## Implementation Details
The barrel pickling process will still preserve vegetables for longer, and doesn't require such a massive investment. I thought this would be a cool reward food for people willing to create two (or three!) different vegetable farms, a spice greenhouse, and automated vinegar.

The fourth vegetable option is entirely for flavor.

This has the #tfc:sealed_jar tag, but you don't have to unseal it to use it in a sandwich. You can clean it out to get your jar back when it's rotten, though.

This would have #tfc:foods/vegetables, but it would allow you to use pickles to make pickles, which could very slowly make an uber-pickle with infinite vegetable nutrient.

## Outcome
Pickle jars can be made in the food processor using fresh cucumbers, garlic, spices, salt, and vinegar. You waste a lot of calories and nutrients in the process, but end up with an easily portable, shelf-stable pickle jar that is a little bit more nutritious than a single vegetable.

image

image


Froffy025:
pickle jars last ONE! whole day longer than jam jars >p<

TomPlop:
I really like the idea but I don't how I feel about it not requiring a time of conservation or being able to eat the glass jar

Redeix:
I thought we said that we didn't need this since we already have pickling. I guess we can accept it since you already put in the work but if you make more prs for stuff we already decline then I won't make more exceptions. I don't mind the concept but it needs some changes:
- I don't think anyone is really going to waste so many ingredients for only a 30% modifier. I would increase that a little or require less salt or remove the garlic.
- The pickles are placed in jars but using them in sandwich recipes for example is just going to delete the jar. No one is going to use it then. You would have to rework it for dedicated sandwich and burger recipes like how jam is done. And I would remove the useable in soup part.
- Where is the texture from? I feel like I've seen it before but maybe in crazy.

Froffy025:
Sorry, I took what you said in the discord as this not being a priority, not as rejecting the concept. We don't have food processor pickling recipes, only brining.

I tried to balance it to what I'd personally find acceptable for late LV. Which 30% modifier do you mean, exactly? I use it in both nutrients and saturation to take the 4 vegetable inputs down to just a little over the value of one.

The usable in soup part is for our polish friends, in case they want zupa ogórkowa. Can't they have some love? >w<

I made the jar using the palette from the new olivine glass batch texture, and the lid is taken directly from the new jar lid texture.

Redeix:
- The modifier for nutrients mostly
- If its useable in soups then there is no way to return the jar

Froffy025:
i've bumped it up to 0.4 nutrient & saturation modifiers and increased the preservation quality a decent bit
image

TomPlop:
What about the others Redeix inquiries?

Froffy025:
I removed the #usable_in_... tags, and added the dedicated recipes requested (plus pickle soup, food processor exclusive.) I added tfc:empty_jar as the craftingRemainder to work for crafting table recipes, and any machine recipe using pickles now returns the glass jars. I think that's everything.

TomPlop:
@Redeix what do you want to do with that PR?

Redeix:
Sorry, forgot to say. I'll be working on more food stuff soon. I want to test this and include it with the other batch of food

PR #4806, opened 2026-08-20
Dye Hammering Recipe

### LINKED ISSUE #4703: [💡]: QoL - Pre-quern Coloured Vessels

- [x] No, it was not.

### A list of popular features/mods the team has already considered and will not implement.

- [x] I understand :(

### What would you like to be added?

Being able to craft coloured vessels in the stone age without needing to craft a quern for dyes.

For example an inefficient but early game recipe to craft an unfired clay vessel with multiple flowers to get a dyed unfired vessel.
* For example: 1 Unfired Vessel + 4-8 `#tfc:makes_orange_dye` flowers = 1 Orange Unfired Vessel.

The items this would include are: Vessel and the Large Vessel

### Why is this needed?

By the time you have copper tools and a quern set up, you are most likely to already have made more than enough vessels to help you on your journey (or have already settled down) and have the tools to create chests to solve your inventory management problems. So you are a lot less likely to make more vessels and invest the time into dying them. Which is a shame because the coloured vessels are very pretty and deserve to be used by everyone from the start.

If there was a way to get dyed item vessels in the stone age people could immediately colour code them to improve the early game inventory management in the stone age, instead of having to guess which brown vessel holds your food over the brown vessels.

This is just to get some QoL into the early game TFC part of the modpack.

jurjen909:
## What is the new behavior?
Flowers and dye producing plants can be crafted with any hammer (it gets damaged) to create a single dye item. Originating from #4703 

## Outcome
Adds new recipes to allow flowers and other dye producing plants to be hammered in a shapeless recipe into dyes.

## Additional Information
- Closes #4703


BlueBoat29:
Would mortar and pestle be better?

Robak132:
Mortar is gated by bronze, this solution is pre-bronze

TomPlop:
The Quern is unlocked right at the start so I'm missing the point here?

Jeuvke:
> The Quern is unlocked right at the start so I'm missing the point here?

Most people make all their vessels before making a quern, disallowing them from using dyes. This would allow people to make coloured vessels right from the start.

BlueBoat29:
We don’t have a stone mortar and pestle? All the ones I’ve ever used are stone

TomPlop:
> We don’t have a stone mortar and pestle? All the ones I’ve ever used are stone

It's a GT tools that use metal. Bronze is the first one though if we really wanted we could add copper but I'm not sure this is an actual issue anyway.

It feels nice to have to make a Quern to unlock dyes and if you really want it takes 10 minutes basically after you launched the game.

BlueBoat29:
stone mortar and pestle that only crushes dyes :>

jurjen909:
The reason I see for wanting this QoL change is to allow people to make colour-coded vessels right from the start in the stone age.

Right now you have to do two whole clay firings: one for your initial clay batch (jug, molds, and all the vessels you want), and the second one for melting your copper nuggets into a copper saw and chisel. Where you then have to gather the raw and smooth stones and combine them into a quern. 
By the time these two firings have been completed you are more than likely going to have made all of the vessels for your journey onwards to best make use of your time when firing clay, so you don't really have the chance to actually make use of the dyes for the vessels unless you specifically under-produce vessels and settle down to wait for a quern. (But that is already too late for vessels in my eyes)

Having an (inefficient) recipe for making dyes in the stone age allows people to express themselves with how they want their vessels to look like, and immensely helps with the inventory management of the early game. Because otherwise you're going to have to look through 10 different brown tinted vessels where your specific items are.


The recipe can be changed to increase the ratio of flowers to dye to create more incentive on using the quern for a larger batch of dyes. But this recipe is more than enough for a quick dye when you literally have no other means of producing it, and want it for something you aren't going to make after your first tools.

Pyritie:
stone hammer is also less efficient than the quern

Pyritie (approved):

TomPlop:
Still feels really weird this isn't a mortar but well for the 2 players that literally can't play without ALL of their vessels being dye I guess this weirdness doesn't matter

TomPlop (approved):

PR #4882, opened 2026-08-31
ae2 proposals

Pyritie:
- Changed all of the ME assembler cleanroom recipes to require oxygenation instead
- Left the cleanroom recipes for the assembler/packer alone, except for the 1ks
- Changed interfaces to use motors instead of conveyors, and changed patprovs to use conveyors instead of robot arms, which should help with the price complaints (left extended and oversize interfaces alone)
- Removed the extra plates from interfaces and patprovs since you already have the crates being the bigger metal material cost (as they're already 8 ingots each)

### DIFF
--- a/kubejs/server_scripts/railways/recipes.js
+++ b/kubejs/server_scripts/railways/recipes.js
@@ -139,13 +139,13 @@ const registerRailWaysRecipes = (event) => {
 		'AAA',
 		'ACA'
 		], {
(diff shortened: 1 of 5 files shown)


kubejs/server_scripts/railways/recipes.js
Pyritie, line 1: oops don't know why this is here

PR #4931, opened 2026-09-05
Oxy Rework

ashleney:
https://github.com/TerraFirmaGreg-Team/Core-Modern/pull/572


kubejs/server_scripts/macaws_for_tfc/tags.js
```diff
@@ -115,4 +115,90 @@ function registerMacawsForTFCItemTags(event) {
 			event.add('mcw_tfc_aio:stripped_furniture', `mcw_tfc_aio:furniture/${wood}_furniture/stripped_${wood}${type}`)
 		})
 	})
+}
+
+function registerMacawsBlockTags(event) {
+    addAll(event, 'mcw_tfc_aio:attic_roofs', [
```
Pyritie: you can just do `event.add(tag, [])` without needing a helper method

kubejs/server_scripts/tfg/machines/recipes.multiblocks.js
```diff
@@ -403,5 +403,40 @@ function registerTFGMultiblockRecipes(event) {
 		.circuit(1)
 		.EUt(GTValues.VA[GTValues.LV])
 		.addMaterialInfo(true)
+
+	// Heat Pump
+	event.recipes.gtceu.shaped('tfg:heat_pump', [
```
Pyritie: both the filter casings and steel pipe casings combined are really expensive. Could one of them be replaced with something else? (also, why hv circuits?)

Since the nano suit lets you ignore temperature (for now), imo this would be better if it was either late LV or early MV

Pyritie: Also I was thinking of reusing one of the refrigerant chemicals that the refrigerator recipes use

kubejs/server_scripts/tfg/machines/recipes.multiblocks.js
```diff
@@ -403,5 +403,40 @@ function registerTFGMultiblockRecipes(event) {
...
+		'BCB',
+		'ABA'
+	], {
+		A: 'gtceu:hv_electric_pump',
+		B: 'gtceu:filter_casing',
+		C: 'gtceu:hv_machine_hull'
+	}).addMaterialInfo().id('tfg:shaped/oxygen_distributor')
+
+	// Higgs Emitter
+	event.recipes.gtceu.shaped('tfg:higgs_emitter', [
```
Pyritie, line 433: the ad astra one was EV

Pyritie (commented):
Would also be good to remove crafting recipes for the ad astra air distributor/gravity normalizer, and add conversion recipes for the new blocks.

Also, what kind of air does the heat pump need as input?

kubejs/server_scripts/tfg/machines/recipes.multiblocks.js
ashleney: as in have the multiblock be made in an assembler? I'm not too big of a fan of that personally

Pyritie: I mean fridges do already have an assembler recipe so it wouldn't be the first time

TomPlop: There is quite a lot of controllers that are made in an Assembler so it's not something that matters

Pyritie (approved):

TomPlop (approved):
Feel like we should merge that one anyway and play around with it when it's in dev. Not like we are pushing an update in a rush soon.

PR #4945, opened 2026-09-08
Changed via romana recipe to use any black dye

Robak132:
## What is the new behavior?
Changed via romana recipe to use any black dye. Currently you have to use inc sac for that.

None, that's just 1 recipe

### DIFF
--- a/kubejs/server_scripts/via_romana/recipes.js
+++ b/kubejs/server_scripts/via_romana/recipes.js
@@ -2,14 +2,14 @@
 "use strict";
 
 const registerViaRomanaRecipes = (event) => {
-	if (!Platform.isLoaded('via_romana')) return
+  if (!Platform.isLoaded('via_romana')) return
 
-	event.remove({ mod: 'via_romana' })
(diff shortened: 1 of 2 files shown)


kubejs/server_scripts/via_romana/recipes.js
```diff
@@ -0,0 +1,15 @@
...
+const registerViaRomanaRecipes = (event) => {
+	if (!Platform.isLoaded('via_romana')) return
+
+	event.remove({ mod: 'via_romana' })
+
+	event.shapeless('via_romana:charting_map', [
+		'minecraft:map',
+		'#forge:dyes/black',
+		'minecraft:feather',
+		'minecraft:string'
```
Pyritie: you'll probably want the #forge:string tag so it works with wool yarn, linen thread, etc

Robak132: Should be good now

Pyritie (approved):

Redeix (approved):

PR #4997, opened 2026-09-17
Make GT covers easier to make 

ashleney:
basically just makes all covers be a shaped crafting recipe, remove the required soldering alloy and no longer require ender dust for voiding covers. 

https://github.com/user-attachments/assets/db025b16-5006-4503-8fa3-0a3b0f712e45


Redeix:
Tbh this doesnt even really feel cheaper. I dont get why covers are even so expensive? If it was up to me I would just have like two tiers of covers and you can right click them in the air to select what type of cover they are.

ashleney:
oh sorry the commit is kinda misleading, it's more about making them less annoying to craft. We could of course outright make them cheaper too if that's desirable, I just wanted to keep some kind of parity with the original gt recipes.

Redeix:
I would definitely push for remaking covers. But I have a feeling someone may fight me on that. Another example is filters. I think most logistics stuff is just way to expensive compared to how often you use them and how other cheap they are in like every other mod

Pyritie:
filters are pretty cheap, they're just one plate and then a bunch of zinc foil which everyone has sitting around at this point. I did also just add a bunch more gem plates to filter crafting too

Redeix:
> filters are pretty cheap, they're just one plate and then a bunch of zinc foil which everyone has sitting around at this point. I did also just add a bunch more gem plates to filter crafting too

I disagree that steel/ gem plates with a bunch of zinc foil is cheap. And the smart filters need circuits. Like its literally just a filter

Pyritie:
how do you feel about create's filters?

Redeix:
> how do you feel about create's filters?

Create filters are much better imo. Having to use cloth is a bit annoying but thats just because currently we dont have a good replacement for the loom until LV. But they do have the advantage of not consuming the filter when nesting. I think GT filters would be better made out of paper and foils.

Redeix:
If you want to keep gems then thats okay but maybe just regular gems instead of plates

Pyritie:
what do you mean consuming the filter?

I do kinda like that the filters are a gem sink, since they're otherwise not used for all that much. The plates could be replaced with normal gems though, I'm not sure why that's a thing

Redeix:
> what do you mean consuming the filter?

When you put a create filter inside another filter it wont consume it. This doesnt really apply to GT though

Pyritie:
it shouldn't consume it, but I guess a filter can contain another filter? unless you mean other covers containing filters?

Redeix:
> it shouldn't consume it, but I guess a filter can contain another filter? unless you mean other covers containing filters?

Just ignore me, I was talking about Create

TomPlop:
If this PR job is to make the cover smoother to craft, it totally does the job. But yeah I don't think it makes them cheaper (it's not needed).

The other discussion is do we want to unlock covers earlier. There is two aspects that make me wonder about it

-the first one is obviously the great feeling you get from progression by unlocking new tools for automation. It's a design that is quite present in TFG, where a lot of QoL tools can be gated much further than other packs

-the second one is how you tell a player that you don't need to start right from the start with crazy automation system but have to go progress a bit before starting them. In a way it either engages a lot of creativity or tell the player they have to do things manually.

Btw regarding this PR I think it would be nice to at least propose it to GTm. We don't do it often and it doesn't cost much. Same thing for the change we did on the Oil texture btw.

Pyritie:
I think gregtech covers should cost similar amounts and be unlocked at similar times to their create equivalents

TomPlop:
I think the filters are okay, you can craft them very early though having to look for a gem is a bit of a bother in game. It does feel cool to have to look for new stuffs exploration and all.

All the other covers sounds really fine to me with two exception. The Machine Cover Controller that could be early on and the Void Cover that I would have love to unlock incredibly early rather that adding the ULV Super Tank/Chest. But I guess it's too late for that.

ashleney:
the way it's in this pr, void covers are available anytime in the metallurgy age but are kinda annoying to make because of the plates. I wonder if we should change plates to rods or something so it's easier on pre-ulv players.

TomPlop:
> the way it's in this pr, void covers are available anytime in the metallurgy age but are kinda annoying to make because of the plates. I wonder if we should change plates to rods or something so it's easier on pre-ulv players.

Pre ULV players don't need them honestly

Pyritie:
hmm maybe foils and then add smithing recipes for ingots directly to foils?

TomPlop:
We are going a bit out of scope I don't think it's needed honestly

Pyritie:
or cloth again, to keep with create

TomPlop:
Are we still speaking about the filters?

TomPlop:
Shouldn't this PR be merged at least? I think it's a good first step anyway

Pyritie:
it's still a draft

TomPlop:
I think @ashleney put it to draft in case we wanted to change some stuffs. But it looks like it's in a good state to me.

ashleney:
yeah as tom said

TomPlop:
So the question is just about filters then we merge? I think the filters even if a bit bothering very early on are quite fine. Not exactly like you need a Gregtech filter before LV anyway

ashleney:
the new filter change made them more bearable I think but I personally really wouldn't mind like 8x'ing how many filters you get per craft. Should probably be part of another PR though

TomPlop:
2 zinc ingots and 1 steel ingots is already so insanely cheap I don't understand honestly, this doesn't feel like needed. Also could you propose this PR to GTm just in case please 🙏

ashleney:
as in, ping gustavo?

Redeix:
Yeah I'm cool if we want to merge this. We can always make another PR later to address other changes we decide on

Pyritie (approved):

